PR #131, opened 2020-10-23
expose basic modular list command

LukeSheard:
Expose a basic modular list command - for now just the same as ``ls`` with a pretty boring implementation as there's no support for anything outside of ``packages/*`` but could be extended later on.

### DIFF
--- a/packages/modular-scripts/src/cli.ts
+++ b/packages/modular-scripts/src/cli.ts
@@ -128,18 +128,34 @@ function run() {
   }
 }
 
+interface WorkspaceMeta {
+  location: string;
+  workspaceDependencies: string[];
+  mismatchedWorkspaceDependencies: string[];
+}
+
 function listPackages(typeArg: string | void) {
   const modularRoot = getModularRoot();
-  const isCorrectType: (dir: string) => boolean = typeArg
-    ? (dir: string) => isModularType(dir, typeArg as PackageType)
-    : isModularPackage;
 
-  const packages = fs
-    .readdirSync(path.join(modularRoot, 'packages'))
-    .filter((dir) =>
-      fs.lstatSync(path.join(modularRoot, 'packages', dir)).isDirectory(),
-    )
-    .filter((dir) => isCorrectType(path.join(modularRoot, 'packages', dir)));
+  const { stdout } = execa.sync('yarnpkg', ['workspaces', 'info'], {
+    cwd: modularRoot,
+  });
+  const workspaceInfo: Record<string, WorkspaceMeta> = JSON.parse(
+    stdout,
+  ) as Record<string, WorkspaceMeta>;
+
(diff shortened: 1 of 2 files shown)


packages/modular-scripts/src/cli.ts
```diff
@@ -118,6 +128,38 @@ function run() {
   }
 }
 
+interface WorkspaceMeta {
+  location: string;
+  workspaceDependencies: string[];
+  mismatchedWorkspaceDependencies: string[];
+}
+
+function listPackages(typeArg: string | void) {
```
NMinhNguyen, line 137: I don't think that `| void` is correct (although I can see it used below - I don't think that's right though). `void` is usually used for functions that don't return anything. Did you mean this?

```suggestion
function listPackages(typeArg?: string) {
```

packages/modular-scripts/src/cli.ts
```diff
@@ -118,6 +128,38 @@ function run() {
...
+  const modularRoot = getModularRoot();
+
+  const { stdout } = execa.sync('yarnpkg', ['workspaces', 'info'], {
+    cwd: modularRoot,
+  });
+  const workspaceInfo: Record<string, WorkspaceMeta> = JSON.parse(
+    stdout,
+  ) as Record<string, WorkspaceMeta>;
```
NMinhNguyen, line 145: Instead of a typecast here, I think it'd be better to use a type assertion (https://github.com/microsoft/TypeScript/pull/32695) to ensure it throws errors (perhaps when using a newer Yarn version?)

packages/modular-scripts/src/cli.ts
```diff
@@ -118,6 +128,38 @@ function run() {
...
+
+  const packages = Object.values(workspaceInfo)
+    .filter((workspaceMeta: WorkspaceMeta) => {
```
NMinhNguyen, line 148: Would this not work?

```suggestion
    .filter((workspaceMeta) => {
```

packages/modular-scripts/src/cli.ts
```diff
@@ -118,6 +128,38 @@ function run() {
...
+      const location = path.join(modularRoot, workspaceMeta.location);
+      if (typeArg) {
+        return isModularType(location, typeArg as PackageType);
+      } else {
+        return isModularPackage(location);
+      }
+    })
+    .map((workspaceMeta: WorkspaceMeta) => {
+      return path.basename(workspaceMeta.location);
+    });
```
NMinhNguyen, line 158: For consistency with the rest of the codebase where we use implicit returns

```suggestion
    .map((workspaceMeta) => path.basename(workspaceMeta.location));
```

NMinhNguyen (changes requested):

LukeSheard:
exploring if we need this - given that `build --all` will land soon.

Issue #125, opened 2020-10-14
The apps created by our project might not work with modular-scripts


threepointone:
We use `create-react-app` to create the app which is aligned to the latest version of react-scripts. However, we use a semver version of `react-scripts` [(code, currently ^3.4.1)](https://github.com/jpmorganchase/modular/blob/master/packages/modular-scripts/package.json#L27) to run the commands. Consider now when 4.0.0 releases, but we haven't updated `modular` yet. We will create an app aligned to v4. If there are any breaking versions in `react-scripts` v4 and/or the structure of the app created, we'd be generating a broken version of the repository (While this may not be true for v4, it may happen in the future anytime, even for dot releases)

A probable fix: we should use `create-react-app@<version>` when installing apps, where `<version>` is hardcoded inside create-modular-react-app/modular-scripts releases.

Issue #120, opened 2020-10-01
Self-hosting the project / dog-fooding


sebinsua:
We should [self-host the project](https://en.wikipedia.org/wiki/Self-hosting_(compilers)) and generally dog-food any code that we write within the project.

The philosophical reason for this would be to _"taste one's own medicine"_ but the practical reasons are much stronger:

- The only way we have to test everything right now is to publish it and then consume the packages as part of an E2E process. This is a very, very slow feedback loop -- and more fragile than I'd like on environment differences or incorrectly written tests, etc.
  - If you're writing test configuration or changing lint rules, we should be able to test these with our own `yarn test` or `yarn lint` command in only a few seconds, for example.
  - There should be a way to quickly create a Modular project and app from within this repository. We have all of the logic required to do so within the tests ([here](https://github.com/jpmorganchase/modular/blob/master/e2e-tests/core.test.ts#L24-L115) and [here](https://github.com/jpmorganchase/modular/blob/4ab966e38e50256ada5f2f7e7832ad80358a6e7b/e2e-tests/core.test.ts#L319-L332)). There's nothing stopping us from pulling this out into a script that a developer can quickly run within the project.
    - Potentially we could also have a `packages/app` (marked as `"private": true`) within the project which would mean people could test changes to `modular-scripts` without any tiresome initialisation of a new project.
- Actually, the E2E process itself is slow because there are multiple `yarn install`s involved in executing `create-modular-react-app`. Elsewhere we talk about caching `yarn install`s but we don't do that here which is why these tests take 6+ minutes to run, instead of taking < 1 minute.
  - If we have a performance problem within the project, it makes sense that we should use [whatever techniques are appropriate](https://github.com/jpmorganchase/modular/issues/62) to fix the problem. That would also be a good test that our tools and instructions related to this are good.
- It's quite weird that we're using different Jest/Lint configurations [in this project](https://github.com/jpmorganchase/modular/blob/master/.eslintrc.js) than the ones we use in [the output project](https://github.com/jpmorganchase/modular/blob/master/packages/eslint-config-modular-app/index.js). I think this was accidental and we should fix it.
- We have [dependencies within our root `package.json` which should actually belong to our packages](https://github.com/jpmorganchase/modular/issues/114).

This doesn't need to be done immediately: I'm just getting my thoughts down.

threepointone:
Made some progress on this with https://github.com/jpmorganchase/modular/pull/132.

Issue #59, opened 2020-08-30
export modular-views.macro


threepointone:
The current implementation exposes the internals of modular, and requires installing codegen.macro. This should probably be inside a module. Further, following https://github.com/jpmorganchase/modular/issues/49, it'll probably have some 'proprietary' logic, and will be hard to change across releases if we don't encapsulate it. 

Opening this issue as a stub, will decide an implementation after #49 lands.

Issue #54, opened 2020-08-19
Debug and resolve why the E2E tests need `--forceExit` when run in a Docker container (or on CI)


sebinsua:
We should debug and resolve why the E2E tests need `--forceExit` when run in a Docker container (or on CI).

See the following comment for further information: https://github.com/jpmorganchase/modular/blob/1fd6cb2efe1bbfef5352447aad4761b04e16e9af/e2e-tests/core.test.ts#L154-L168

threepointone:
Marking this as high pri, since it's been marked as skipped in our tests, and we should fix it soon.

threepointone:
This issue isn't particularly accurate anymore since we've changed the way we run the tests and have a fresh set of problems. Closing this issue in favour of https://github.com/jpmorganchase/modular/issues/169.

PR #132, opened 2020-11-08
refactor: update dependencies + self host repository.

threepointone:
This PR is a refactor of the codebase, but without changing (much) functionality. It may seem like a lot, but that's exaggerated by the fact that a lot of it is just moved round. Here are the details - 

- This updates `react-scripts`/`@craco/craco` to support `create-react-app@4.0.0`. As part of that, I've also disabled the in-browser linting experience (https://github.com/gsoft-inc/craco/issues/205). *This is temporary*. I'll enable it when it's fixed on craco's side, but that shouldn't block this PR. This should fix https://github.com/jpmorganchase/modular/issues/125. 
- The refactor is mainly around making this repository self-hosted (https://github.com/jpmorganchase/modular/issues/120). The steps taken to achieve that were: 
  - e2e-tests were split up and inlined into `modular-scripts` / `create-modular-react-app`, as applicable. This also enabled removing verdaccio and associated complexity.
  - `cra-template-modular-typescript` was inlined into `modular-scripts`; this has the benfit of preventing one waterfall step when adding a new app to a project. 
  - We had a bug where we needed at least one app named `app` for all tests to run, I fixed that. We still need at least one app, but it can be named whatever. (I've named it `modular-site`, and we can use it for experiments, and maybe hosting documentation). In the future we dhsould fix it so it doesn't need any at all, but that's an edge case. 
- I've removed `modular-scripts/src/getModularViewMap`, and will instead implement `modular-views.macro` (https://github.com/jpmorganchase/modular/issues/59). I'm working on it right now, but it shouldn't block this PR. 

### caveats 

- I need to implement and write tests for the views macro, as mentioned above. 
- While this runs all tests on CI, it skips one test, the one that starts up a server and tests whether the app works as expected. This is because it still leaves hanging processes, and will eat up our github-actions minutes. Working on fixing it, it's a weird one. (probably related to https://github.com/jpmorganchase/modular/issues/54)
- I haven't verified builds yet, will do so. (EDIT: done, looks fine)
- I need to simplify the root configurations for babel/eslint so that it uses the `modular` configs. It's not critical to this PR so I figured it could wait. 
- all tests take about 550s on my 2015 macbook pro. That kinda sucks. A lot of the time is literally just fetching dependencies and installing. We should setup caches for our build, and further use cached versions of dependencies when installing. This is less of a problem now that the repo is public, but still. (EDIT: now under 4 minutes, not bad.)

I'm looking for a review so I can start making changes in parallel to the above fixes. Have a look!


package.json
```diff
@@ -12,9 +11,8 @@
     "prettier": "prettier --write .",
     "lint": "node ./scripts/lint.js --cache --ext .js,.ts,.tsx --max-warnings 0",
     "linc": "node ./scripts/linc.js --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
-    "test": "jest",
-    "e2e": "jest -c jest-e2e.config.js --runInBand",
-    "e2e:docker": "./e2e-tests/run-tests-in-docker.sh",
+    "modular": "npx ts-node packages/modular-scripts/src/cli.ts",
```
threepointone, line 14: this doesn't appear to run on github actions, hrmm.

packages/create-modular-react-app/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,170 @@
+import createModularApp from '../cli';
+import fs from 'fs-extra';
+import path from 'path';
+import rimraf from 'rimraf';
+import tree from 'tree-view-for-tests';
+
+const repoDirectory = '../../../sample-modular-project';
```
threepointone, line 7: this should be a tmp directory instead, will fix.

packages/modular-scripts/craco.config.js
```diff
@@ -48,10 +82,7 @@ module.exports = {
         ...jestConfig,
         rootDir: absolutePackagesPath,
         roots: ['<rootDir>'],
-        testMatch: [
-          '<rootDir>/*/src/**/__tests__/**/*.{js,ts,tsx}',
```
threepointone, line 52: this was annoyingly calling all files inside __tests_, even helpers.

packages/modular-scripts/src/__tests__/TestApp.test-tsx
```diff
@@ -25,14 +23,8 @@ function App(): JSX.Element {
           Learn React
         </a>
       </header>
-      <div data-testid="views">
```
threepointone, line 28: we'll write tests for the views macro separately.

packages/modular-site/.env
```diff
@@ -0,0 +1 @@
+SKIP_PREFLIGHT_CHECK=true
```
threepointone, line 1: this is required because of a of the existence of multiple versions of babel-jest. will fix.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
-    "module": "esnext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
+    "module": "CommonJS" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
```
threepointone, line 7: changed this to CommonJS so I can use ts-node locally. Open to alternative suggestions.

Cono52 (approved):

packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,300 @@
...
+      all: true,
+      reject: false,
+      env: {
+        CI: 'true',
+      },
+    });
+
+    // I think this is wrong, it's not running tests
+    // in sample-package and sample-view
+    expect(censorVariableTimes(output.all)).toMatchInlineSnapshot(`
```
threepointone, line 288: reflects https://github.com/jpmorganchase/modular/issues/130

package.json
threepointone: fixed.

threepointone:
omg tests pass

PR #137, opened 2020-11-13
remove cra-template-modular-typescript from changeset

threepointone:
I'm trying to do a release, and it's failing because it's expecting a build on cra-template-modular-typescript (which was removed in https://github.com/jpmorganchase/modular/pull/132). This PR removes that intention from the one changeset that refers to it.


threepointone:
Landing this.

PR #138, opened 2020-11-13
Fix broken release. 

threepointone:
I broke the build :/ During the refactor, I forgot to had the shebang for the script that gets called when creating a new project. This PR fixes that, and updates hashes in some snapshot tests.


threepointone:
Landing this since it's a bit urgent.

threepointone:
(well, after tests pass)

Issue #130, opened 2020-10-23
modular test script does not run sub-package test


zemeolotu:
## Steps to reproduce

1. Create a new app - `yarn create modular-react-app test-app`
2. Navigate into new app directory - `cd test-app`
3. Add a new React widget sub package - `yarn modular add sub-app`
4. Run test - `yarn test`

## Expected Result:
Run both tests in `app` and `sub-app` directories.

## Actual Result:
Only runs tests in `app` sub-package

## Version
create-modular-react-app@0.3.0

threepointone:
Ok, I figured out what the problem was here. Fix and longer explanation in https://github.com/jpmorganchase/modular/pull/141, tl;dr for an immediate fix -> move your source code inside a folder called `src` in the package/view, and update `main` inside `package.json` to point to the root file. 

Will leave this issue open till the PR lands and I do a release.

threepointone:
Leaving this open till I do a release / sync internally.

threepointone:
This is released and synced internally, closing.

PR #141, opened 2020-11-14
run tests correctly for packages and views

threepointone:
FIxes https://github.com/jpmorganchase/modular/issues/130

When running `modular test`, we weren't running tests for views or packages. This was because our `testMatch` config ran against `['<rootDir>/*/src/**/*.{spec,test}.{js,ts,tsx}']`, but the packages we generated with `modular add` weren't putting the source or tests inside an `src` folder (and using the root of the package instead). This PR fixes the template used to generate the packages, and updates tests to verify this. Existing projects can 'fix' their problem by moving their code and tests inside a 'src' directory, and marking 'main' in their package.json to point to `src/index.ts`/`src/index.tsx`/whatever their root file is.


packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -287,17 +285,23 @@ describe('modular-scripts', () => {
...
+    // removed those tokens from the string for the snapshot test.
+    // Open to suggestions/fixes.
+
+    // eslint-disable-next-line no-control-regex
+    expect(censorVariableTimes(output.all?.replace(/|\[\d+./gm, '')))
+      .toMatchInlineSnapshot(`
       "$ ts-node packages/modular-scripts/src/cli.ts test sample-app sample-package sample-view
-      PASS ../sample-app/src/__tests__/App.test.tsx
-        ✓ renders learn react link (? ms)
+       PASS  ../sample-package/src/__tests__/index.test.ts
```
threepointone, line 297: The tests are failing because the order of these 3 tests is not deterministic. I'm not even sure how to begin to solve that, so I'm leaving this as is for now while I think about it. Everything else seems a-ok tho.

threepointone:
Hrmm, facing hanging tasks again in the now enabled test, hence CI failing. That said, It's still exiting and not eating any extra minutes, so I think it's fine... shouldn't block review imo.

threepointone:
I've disabled the test in CI till I can figure it out.

packages/modular-scripts/types/package/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "PackageName__",
   "version": "1.0.0",
-  "main": "index.ts",
+  "main": "src/index.ts",
```
matt-d-webb, line 4: Would removing `*/src/**/*` from the RegEx not fix this also? Or do we need a point of reference `src` to avoid running unexpected tests?

threepointone: We could... but that would mean we have to change our typescript matches as well. Which could be fine, but that'll complicate making builds later, because it'll be hard to isolate built artifacts. I think it's better to isolate this stuff in a folder.

threepointone:
I'm going to land this since it's a bug fix/patch release, we can discuss whether we want to change the match paths in a later revision. Thanks!

Issue #60, opened 2020-08-31
export modular.tsconfig


threepointone:
We plop a tsconfig.json into the generated repository right now, but inline all the settings. This will make it harder to change if and when we make changes in the future (I can already think of a couple of changes I'd like to make). Instead, we should make a tsconfig.json (maybe as a package? dunno) and then in the created file, use `extends: "modular.tsconfig"`.

threepointone:
leaving this open till it's integrated properly

threepointone:
finished in https://github.com/jpmorganchase/modular/pulls?q=is%3Apr+is%3Aclosed

PR #142, opened 2020-11-14
modular.tsconfig

threepointone:
Fixes https://github.com/jpmorganchase/modular/issues/60

This creates a new package, `modular.tsconfig`, for usage with modular projects. We can iterate on this, and further improve it later. The benefit is that we can make meaningful changes and propagate them to consumers.

This PR does _not_ make changes to `create-modular-react-app` to use this just yet, because we'll have to publish this first before we can use it in tests, so I've punted on that for the next PR.

(PS: Much like jest/babel/etc configs, would've been nice if we could hide this and prevent changes altogether, but I don't think the community/ecosystem is mature enough for that yet. We'll tackle that at some point)

### DIFF
--- a/.changeset/early-rings-sing.md
+++ b/.changeset/early-rings-sing.md
@@ -1,0 +1,5 @@
+---
+'modular.tsconfig': minor
+---
+
+Create modular.tsconfig, a standard tsconfig for modular project
--- a/packages/modular.tsconfig/README.md
+++ b/packages/modular.tsconfig/README.md
@@ -1,0 +1,6 @@
+## modular.tsconfig
+
+This is a standard
+[tsconfig](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)
+package, made for usage by [modular](https://github.com/jpmorganchase/modular)
+projects.
--- a/packages/modular.tsconfig/modular.tsconfig.json
+++ b/packages/modular.tsconfig/modular.tsconfig.json
@@ -1,0 +1,18 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["dom", "dom.iterable", "esnext"],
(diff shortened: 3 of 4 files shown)


LukeSheard:
> Much like jest/babel/etc configs, would've been nice if we could hide this and prevent changes altogether

Why not? I think this is a serious discussion we should have before merging this. Especially given that we're not beginning to start iterating on this. From the discussions we've had we know that once we make things too loose and "easy" there'll be no turning back. Maybe now this PR is where we start locking this stuff down so to speak.

threepointone:
Yeah we can discuss this, it’s not super critical to land this just yet. Some quick notes: 
- firstly, we’ll have fairly strict defaults, not a problem. 
- but for organisations just starting the typescript journey, unless they have experts on their team, they would be wary of not having some flexibility while they become comfortable using it. Given an option of strict or nothing, they’ll choose nothing (ie- they won’t even use modular). Or they’ll just write .js files, bypassing typescript altogether.
- it’s another thing for teams that are already familiar with typescript; in such a case, it’s possible to put _another_ layer of strictness with other tooling (or simply, reject any PRs/changes to tsconfig.json).
- finally, I’m not sure it’s possible to configure a custom path for this configuration? Which means it’s not possible to hide it I think. Happy to be wrong, but the above points are still legit I think. 

Happy to discuss JP specific concerns in person too.

matt-d-webb (approved):

threepointone:
Not going to merge this just yet, I'll do it after I release what we have in master, then work on integrating this into create-modular-react-app's install flow (and our own repo, ofc), and then release it. Tentatively this weekend, maybe early next week.

threepointone:
Going to land this so I can get to work integrating it with the rest of the codebase/projects.

PR #145, opened 2020-11-15
add CHANGELOG.md to defaultHashIgnores in tree-view-for-tests

threepointone:
Much like we ignore the hash for package.json when generating a tree in snapshot tests, this PR adds CHANGELOG.md as well to the ignore list. Otherwise, every time we do a release, the hash on the changelog changes and breaks our tests/builds.


threepointone:
This doesn't need a changeset since it never gets published. Going to land this once tests pass since it'll unblock the other PRs in flight.

PR #153, opened 2020-11-19
add integrated jest-junit story for reporting into Jenkins

LukeSheard:
Given the heavy usage of Jenkins within in the firm it makes sense to have an out the box setup for junit integration and reporting. This adds a default reporter into the test environment.

### DIFF
--- a/packages/create-modular-react-app/template/gitignore
+++ b/packages/create-modular-react-app/template/gitignore
@@ -7,6 +7,7 @@ node_modules
 
 # testing
 /coverage
+/target
 
 # production
 /packages/*/build
--- a/packages/modular-scripts/craco.config.js
+++ b/packages/modular-scripts/craco.config.js
@@ -80,6 +80,15 @@ module.exports = {
 
       return {
         ...jestConfig,
+        reporters: [
+          'default',
+          [
+            'jest-junit',
+            {
+              outputDirectory: path.resolve(modularRoot, 'target'),
+            },
+          ],
+        ],
         rootDir: absolutePackagesPath,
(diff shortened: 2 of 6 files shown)


packages/create-modular-react-app/template/gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 
 # testing
 /coverage
+/target
```
threepointone, line 10: "target" is an odd name for this (I assume there's some historical reason for it). Can it go into /coverage?

threepointone (commented):
Broadly looks good to me. Afk, so I'll stamp it on the morning after trying it out once. Thanks!

packages/create-modular-react-app/template/gitignore
LukeSheard: Historical - since it's related to  [maven](https://maven.apache.org/surefire/maven-surefire-report-plugin/)

> The Surefire Report Plugin parses the generated TEST-*.xml files under ${basedir}/target/surefire-reports

packages/modular-scripts/craco.config.js
```diff
@@ -80,6 +80,15 @@ module.exports = {
 
       return {
         ...jestConfig,
+        reporters: [
+          'default',
+          [
+            'jest-junit',
```
threepointone, line 86: I agree this feature would valuable to include, but thinking about whether it should happen by default. How would you feel about putting this behind a flag? Say, `--coverage-junit` or something.

threepointone:
You can run `yarn test -u` to update snapshot tests to pass CI.

Issue #112, opened 2020-09-24
Tests centre around `packages/app` even though they are run for the whole of `packages/*`


sebinsua:
<img width="386" alt="Screenshot 2020-09-24 at 12 58 31" src="https://user-images.githubusercontent.com/152098/94142206-ce07ef80-fe65-11ea-9ac8-98e2e6494783.png">

As you can see from the `../app-two/src/` above, the paths seem to be relative to `packages/app`.

I'm assuming this is because of the current working directory, so perhaps this should be changed? However I have a memory that `react-scripts` and `craco` were quite fussy about the directory they were run from so we will need to figure out how to do this.

Functionally, I don't think it presents a huge issue, but aesthetically it's gross. ~~(Actually, perhaps it might impact on the coverage, in which case it would be a functional concern...)~~

threepointone:
Fixed in https://github.com/jpmorganchase/modular/pull/155

PR #155, opened 2020-11-20
migrate test command to jest directly

LukeSheard:
Fix for #112 where currently we require that an `app` be available in the modular structure to be able to run tests. This is because we leverage `craco` and therefore need there to be a `create-react-app` package available. 

This PR is the first step in migrating directly to `jest` for `modular test`. I've exposed `--watchAll` and `--coverage` for now, but we should look to work out exactly what abstraction we want to create on top of jest. This also will allow us to leverage incremental testing in future.

### DIFF
--- a/.changeset/witty-foxes-punch.md
+++ b/.changeset/witty-foxes-punch.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': patch
+---
+
+Run jest manually, instead of using react-scripts to do so.
--- a/packages/modular-scripts/craco.config.js
+++ b/packages/modular-scripts/craco.config.js
@@ -85,6 +85,9 @@ module.exports = {
         testMatch: ['<rootDir>/*/src/**/*.{spec,test}.{js,ts,tsx}'],
         coverageDirectory: path.resolve(modularRoot, 'coverage'),
         collectCoverageFrom: ['<rootDir>/*/src/**/*.{js,ts,tsx}', '!**/*.d.ts'],
+        setupFiles: jestConfig.setupFiles.concat([
+          require.resolve('./jest-setupFiles.js'),
+        ]),
         setupFilesAfterEnv: glob.sync(
           `${absoluteModularGlobalConfigsPath}/setupTests.{js,ts,tsx}`,
           {
--- a/packages/modular-scripts/craco.config.js
+++ b/packages/modular-scripts/craco.config.js
@@ -86,7 +86,7 @@ module.exports = {
         coverageDirectory: path.resolve(modularRoot, 'coverage'),
         collectCoverageFrom: ['<rootDir>/*/src/**/*.{js,ts,tsx}', '!**/*.d.ts'],
         setupFiles: jestConfig.setupFiles.concat([
-          require.resolve('./jest-setupFiles.js'),
+          path.join(__dirname, './jest-setupFiles.js'),
         ]),
         setupFilesAfterEnv: glob.sync(
           `${absoluteModularGlobalConfigsPath}/setupTests.{js,ts,tsx}`,
--- a/packages/modular-scripts/src/__tests__/index.test.ts
+++ b/packages/modular-scripts/src/__tests__/index.test.ts
@@ -42,6 +42,7 @@ function getBrowser() {
 function modular(str: string, opts: Record<string, unknown> = {}) {
   return execa('yarnpkg', ['modular', ...str.split(' ')], {
     cwd: modularRoot,
+    cleanup: true,
     ...opts,
   });
 }
@@ -151,7 +152,7 @@ afterAll(() => {
   rimraf.sync(path.join(packagesPath, 'sample-view'));
   rimraf.sync(path.join(packagesPath, 'sample-package'));
   // run yarn so yarn.lock gets reset
-  return execa('yarnpkg', [], {
+  return execa.sync('yarnpkg', [], {
     cwd: modularRoot,
   });
 });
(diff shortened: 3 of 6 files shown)


threepointone:
This is looking good, but please fill in a description with motivation for this change.  You should still pass MODULAR_ROOT in env when calling this (or setup a workaround). yarn prettier and yarn test -u to pass CI too. Thanks!

LukeSheard:
@threepointone added a better description and also abstracted jest slightly so that I didn't need to deal with the imports. Tests were failing though, so not sure what's going on there.

packages/modular-scripts/src/cli.ts
```diff
@@ -226,27 +227,25 @@ function addApp(name: string) {
...
   }
 
-  return execSync(cracoBin, ['test', '--config', cracoConfig, ...args], {
-    cwd: path.join(modularRoot, 'packages', appDirent.name),
+  return execSync(jestBin, argv, {
+    cwd: modularRoot,
     log: false,
     // @ts-ignore
     env: {
+      BABEL_ENV: 'test',
```
threepointone, line 246: Does CRA add these vars too?

packages/modular-scripts/modular.jest.config.js
```diff
@@ -0,0 +1,4 @@
+const { createJestConfig } = require('@craco/craco');
```
threepointone, line 1: could you explain why this is in a separate file? (And not under src?)

threepointone (commented):
The failing test (https://github.com/jpmorganchase/modular/runs/1436116049?check_suite_focus=true) might have to do with the `--projects` flag, unsure. (https://github.com/facebook/jest/issues/5595 / https://jestjs.io/docs/en/cli#--projects-path1--pathn)

packages/modular-scripts/src/cli.ts
LukeSheard: They're mapped from https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/test.js#L12

threepointone:
I think we may have to copy more stuff from react-scripts (https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/test.js). Specifically, loading their env vars, calling verifyPackageTree, resolveJestDefaultEnvironment. I'll look into it.

threepointone:
Ok, I did those bits too, it's looking pretty decent now. Added a changeset as well. Could you do a review of my changes and we'll merge this in?

.changeset/witty-foxes-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'modular-scripts': patch
```
LukeSheard, line 2: I'd argue this is probably more of a minor than a patch. 🛩️

packages/modular-scripts/craco.config.js
```diff
@@ -85,6 +85,9 @@ module.exports = {
         testMatch: ['<rootDir>/*/src/**/*.{spec,test}.{js,ts,tsx}'],
         coverageDirectory: path.resolve(modularRoot, 'coverage'),
         collectCoverageFrom: ['<rootDir>/*/src/**/*.{js,ts,tsx}', '!**/*.d.ts'],
+        setupFiles: jestConfig.setupFiles.concat([
+          require.resolve('./jest-setupFiles.js'),
```
LukeSheard, line 89: My personal preference would be `path.join(__dirname, ...)`, but I guess the semantics of `require` are pretty baked in now.

.changeset/witty-foxes-punch.md
threepointone: I'm using minors before 1.0 to mean breaking changes, think it would still apply?

(I don't feel super strongly about it, so either works for me.)

packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -288,13 +289,13 @@ describe('modular-scripts', () => {
     const cleanedOutput = output.all?.replace(/|\[\d+./gm, '');
 
     expect(cleanedOutput).toContain(
-      'PASS  ../sample-app/src/__tests__/App.test.tsx',
+      'PASS packages/sample-app/src/__tests__/App.test.tsx',
```
LukeSheard, line 292: Bangin'

packages/modular-scripts/craco.config.js
threepointone: I don't recollect why I did require.resolve here. Will change.

packages/modular-scripts/src/cli.ts
```diff
@@ -39,6 +41,7 @@ function execSync(
     stdin: process.stdin,
     stderr: process.stderr,
     stdout: process.stdout,
+    cleanup: true,
```
LukeSheard, line 44: I guess this is the issue we were talking about when it comes to passing through `SIGTERM` etc. ?

threepointone: Yeah, but even this doesn't really fix it. Still good to have so I left it in.

packages/modular-scripts/src/cli.ts
```diff
@@ -227,27 +230,85 @@ function addApp(name: string) {
...
+  argv.push('--env', testEnvironment || '');
+
+  // ends the section copied from CRA
 
-  return execSync(cracoBin, ['test', '--config', cracoConfig, ...args], {
-    cwd: path.join(modularRoot, 'packages', appDirent.name),
+  return execSync(jestBin, argv, {
+    cwd: modularRoot,
     log: false,
+    cleanup: true,
```
LukeSheard, line 306: This is already implicit from above, no? 

https://github.com/jpmorganchase/modular/pull/155/files#diff-228bd4df6d9d99fd05a68843f3ecd0c638ee9a58849130167f7ddd25a39c6cd2R44

threepointone: yeah, will remove

Issue #156, opened 2020-11-22
creating a new view/package doesn't replace the stub names with the new view/package name 


threepointone:
I noticed today that it leaves behind the "ComponentName__" stubs in the created files. Will fix.

threepointone:
Fixed in #167

Issue #149, opened 2020-11-18
react/react-dom (and other core dependencies) should be installed at the root 


threepointone:
In modular projects, react/react-dom should probably be installed at the root if the project. This also prevents having to specify the package versions in created views ([which are lagging](https://github.com/jpmorganchase/modular/blob/510a8764409553d26b2736bf479585f25aa2e97a/packages/modular-scripts/types/view/package.json#L7-L8) behind the versions that create-react-app installs right now, 17 vs 16)

This also applies to typescript, tho maybe that be a separate PR, and might have a couple of other considerations.

NMinhNguyen:
> This also prevents having to specify the package versions in created views

Would this not be an issue when using Yarn v2 (or any tool like ESLint for example that checks imports and dependencies)? When it comes to publishing packages, they'll need to properly declare their dependencies as well.

threepointone:
Yeah, `modular package` will rewrite package.json to include those dependencies (as well as packages imported from local workspaces) when making build artifacts.

threepointone:
Fixed in #167

PR #167, opened 2020-11-25
fixes for #149 and #156

threepointone:
- installs react and react-dom at the root of the generated repository, keeping it uniform across the project.
- doesn't install react/react-dom when making a new view
- correctly replaces placeholder names when generating new packages.
- additionally, I've disabled the tests that runs tests, because the output is slightly different locally and in CI (the whitespace is different). Will fix later.
- further, this fixes a bug where options weren't being passed on to the test runner.


packages/modular-scripts/src/cli.ts
```diff
@@ -230,10 +245,9 @@ function addApp(name: string) {
 function test(args: string[]) {
   const modularRoot = getModularRoot();
 
-  let argv = mri(process.argv.slice(3))._.concat([
-    '--config',
-    path.join(__dirname, '..', 'jest-config.js'),
-  ]);
+  let argv = process.argv
+    .slice(3)
+    .concat(['--config', path.join(__dirname, '..', 'jest-config.js')]);
```
threepointone, line 250: trying to pass an additional config on top of this refuses to run jest, so this is actually fine and expected behaviour (tho we could add an error message to be nice. maybe later.)

threepointone:
Landing this because it isn't dramatic or risky, has high impact because it fixes bugs, and I need it for some other stuff tonight.

packages/modular-scripts/src/cli.ts
LukeSheard: Let's raise an issue for this? Would be a good "first issue" for someone. 😄

threepointone: I suppose, yeah. It's not supported behaviour in create-react-app either, so we should mirror whatever that behaviour is. I'll file something tomorrow (or I may just fix it).

threepointone: I checked, and create-react-app doesn't do anything differently either, so we shouldn't too. 

But, if (and this is a big if) we do move away from create-react-app, we can and should rethink the experience. 

Marking this conversation as resolved.

PR #168, opened 2020-11-25
chore: Fixes dependency versions as recommended by manypkg.

threepointone:
I ran `npx @manypkg/cli fix`, then committed changes. Easy. 

No changeset needed for this.


threepointone:
Landing this, it's just busywork.

PR #171, opened 2020-11-26
Don't use create-react-app to bootstrap an app

threepointone:
This PR bypasses using create-react-app to bootstrap an application, instead using a simple template to make the workspace. Reasoning:
- We don't need the latest versions of react/react-dom, or any other dependencies, installed in the created workspace, since we install those at the root now (#167, #149)
- We don't even need react-scripts installed there, since modular-scripts already bundles it, and we control the installed version in that manner
- This makes the bootstrapping experience much faster, preventing a number of network fetches

This PR also passes on '--prefer-offline' when passed to yarn create modular-react-app, potentially making the complete install experience a lot quicker.


matt-d-webb (approved):

matt-d-webb:
Makes sense. Appreciated the comment description before reviewing.

PR #176, opened 2020-11-26
delete cra-template-modular-typescript

threepointone:
`cra-template-modular-typescript` is not necessary anymore since https://github.com/jpmorganchase/modular/pull/171/ landed. This PR removes it.

This doesn't need a changeset.


threepointone:
Landing this, just busywork.

PR #185, opened 2020-11-29
Allow multiple references to the views macro

threepointone:
This PR modifies the macro so you can have multiple references to the macro in one file and it won't output the map multiple times (also avoiding the annoying error).

### DIFF
--- a/packages/modular-views.macro/src/index.macro.js
+++ b/packages/modular-views.macro/src/index.macro.js
@@ -56,11 +56,15 @@ const viewsMap = `({
 })`;
 
 function views({ references, babel }) {
+  if (!references.default || references.default.length === 0) {
+    return;
+  }
+
   references.default[0].hub.file.path.node.body.unshift(
     babel.template.ast(`const __views__map__ = ${viewsMap};`),
   );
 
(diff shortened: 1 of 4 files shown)


packages/modular-views.macro/src/index.macro.js
```diff
@@ -56,20 +56,23 @@ const viewsMap = `({
...
-    throw new Error(
-      'Do not use modular-views.macro more than once in a module, it will bloat your bundle size.',
-    );
-    // TODO: would be nice if we could restrict it to only one usage per _program_.
+  if (!references.default || references.default.length === 0) {
+    return;
   }
 
-  references.default.forEach((ref) => ref.replaceWithSourceString(viewsMap));
+  references.default[0].hub.file.path.node.body.unshift(
```
threepointone, line 63: this should probably be inserted after all import statements, not the very top like here.

threepointone: Ok I checked, and it's not a syntax error if this is at the very top, and it'll compile such that import statements get hoisted anyway. Leaving as is.

Issue #169, opened 2020-11-25
Fix tests, but actually this time. 


threepointone:
Our tests behave a little funnily. They got a little better after we moved to a self hosted model, but there are still some rough edges. 

- [x] Run `yarn test`. Wait for a few seconds, then hit ctrl-c to end it. You'll notice that it still runs in the background, creating the sample repos/apps/projects/whatever, all the way to completion. That's weird.  (Fixed in https://github.com/jpmorganchase/modular/issues/169)

- [x] We have the test that starts up the dev server disabled right now https://github.com/jpmorganchase/modular/blob/09b638ad29e81b786465a210b508ee18af67709b/packages/modular-scripts/src/__tests__/index.test.ts#L195 This is because when you run it, and it ends, you'll see that it leaves hanging node process in the background, i.e.: In activity monitor, you'll see the `node` processes still running, and you'll have to kill them manually. This doesn't happen in production, so it's super weird. (Fixed in https://github.com/jpmorganchase/modular/issues/169, enabled in https://github.com/jpmorganchase/modular/pull/189) 

- [x] We have the test that runs tests (meta hehe) disabled right now too https://github.com/jpmorganchase/modular/blob/09b638ad29e81b786465a210b508ee18af67709b/packages/modular-scripts/src/__tests__/index.test.ts#L272 The problem here is that the output of running the tests gives the 'coloured' strings, which it shouldn't when CI: true is passed as an environment variable. So I worked around that by stripping the characters with a regex that up, but the whitespace still appears to be different, leading the out put to *slightly* be different locally/in CI. The correct fix here is probably to generate the text properly in the first place. (Fixed in https://github.com/jpmorganchase/modular/issues/169)

- [x] @Cono52 mentioned that the tests don't pass correctly in windows at all. So that's something else to look into. (Fixed in https://github.com/jpmorganchase/modular/issues/169)

- [x] @Cono52 further mentioned that they don't run inside docker. This isn't super critical since we don't use that workflow anymore, but would be good to investigate why. (Deprioritised since we don't use docker for tests anymore)

threepointone:
My initial guess is that `execa` is kinda buggy, so maybe we can experiment replacing it with something like `cross-spawn` for shelling out. Unsure.

threepointone:
https://github.com/jpmorganchase/modular/issues/169 actually did fix most of these, thanks a bunch @Cono52, that was dope. Closing this issue.

PR #186, opened 2020-12-02
Fix windows tests

Cono52:
The issues I mentioned with ```execa``` were actually related to race conditions between the tests and various resources (i think).
--runInBand as a default resolved that.

Other issues were simply paths + files hashing.

This PR resolves only some of the tasks in #169


packages/create-modular-react-app/src/cli.ts
```diff
@@ -59,7 +59,9 @@ export default function createModularApp(argv: {
   }
 
   const newModularRoot =
-    name[0] === '/' ? /* absolute */ name : path.join(process.cwd(), name);
+    name[0] === '/' || name.includes(':\\')
```
threepointone, line 62: good catch, nice fix!

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
```
threepointone, line 1: great addition, thanks!

threepointone (commented):
Looking good, added some questions

packages/tree-view-for-tests/src/index.ts
```diff
@@ -7,6 +7,16 @@ function times(str: string, length: number) {
   return Array.from({ length }, () => str).join('');
 }
 
+// Depending on the file type we either remove carriage returns or not
+// so we get the same hash on windows or mac.
+function agnosticHash(_path: string, fileString: string): string {
```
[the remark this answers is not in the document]
Cono52: oo ye you're right I could just do it for all files and just update the snapshot.
I think i ended up doing it this way so it would produce the same hashes in the tests as oppose to changing the hashes.

but ye I will remove this logic and update the hashes.

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+    {
+      "name": "Debug Test",
+      "type": "node",
+      "request": "launch",
+      "args": ["${relativeFile}"],
+      "runtimeArgs": [
+        "--nolazy",
+        "-r",
+        "ts-node/register",
+        "packages/modular-scripts/src/cli.ts",
```
threepointone, line 12: Should there be a --runInBand here too?

threepointone (approved):
Left one question, but it looks good to me. Preemptively stamping, thank you!

package.json
```diff
@@ -12,7 +12,7 @@
     "lint": "node ./scripts/lint.js --cache --ext .js,.ts,.tsx --max-warnings 0",
     "linc": "node ./scripts/linc.js --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
     "modular": "ts-node packages/modular-scripts/src/cli.ts",
-    "test": "yarn modular test",
+    "test": "yarn modular test --watchAll=false --runInBand",
```
LukeSheard, line 15: Rather than have this - potentially we just always add ``--runInBand`` into the `test` command when we're running with `CI=true` in the environment?

threepointone: We don't want that for every modular project, I'd think, it'll make test suites slow. We only only want it for our own because we mess with the file system and stuff which breaks snapshot tests because they may not be stable when running them all together.

packages/tree-view-for-tests/src/index.ts
```diff
@@ -43,7 +42,7 @@ function tree(
       `${times('#', level)}${path.basename(_path)}`,
       options.hashIgnores.includes(path.basename(_path))
         ? undefined
-        : `#${hash(fs.readFileSync(_path, 'utf8'))}`,
+        : `#${hash(fs.readFileSync(_path, 'utf8').split('\r').join(''))}`,
```
threepointone, line 45: how about this - 
```suggestion
        : `#${hash(fs.readFileSync(_path, 'utf8').replace(/\r/gm, ''))}`,
```

(haven't tested, just curious)

threepointone:
I think you can enable this test now and it'll pass https://github.com/jpmorganchase/modular/blob/6fdd17120c26e22c1c0e2ef2a79866b9f8295bbe/packages/modular-scripts/src/__tests__/index.test.ts#L272

PR #189, opened 2020-12-07
enable the test that tests the dev server

threepointone:
This PR enables the one disabled test we have, since it doesn't leak since https://github.com/jpmorganchase/modular/issues/169.


threepointone:
Ok fantastic, looks like it works just fine. Merging this in.

PR #190, opened 2020-12-10
Add ability to provide setupFiles to jest from modular configuration

LukeSheard:


threepointone:
Looks good, could you fix the breaking CI build? (I think putting `export {}` should pass it.) Will stamp after CI passes.

threepointone:
This also needs a changeset, probably patch.

packages/tree-view-for-tests/src/__tests__/tree.test.ts
```diff
@@ -12,6 +12,9 @@ test('it can serialise a folder', () => {
     ├─ .npmignore #1rstiru
     ├─ CHANGELOG.md
     ├─ README.md #r0jsfd
+    ├─ build
```
threepointone, line 15: this shouldn't be committed

LukeSheard: Yeah - this is annoying I will see if I can ignore this dir so it doesn't keep happening.

threepointone: we technically shouldn't ever have to run a manual build locally on these packages, were you experimenting with something?

threepointone (approved):
tentatively stamping, but please fix the failing snapshot test.

LukeSheard:
Fixed hash now and changeset done.

PR #192, opened 2020-12-10
move verify packages and rename setupFiles

LukeSheard:
Fix issue where jest workers would exit without logs if there was an issue in the package tree.

### DIFF
--- a/packages/modular-scripts/src/cli.ts
+++ b/packages/modular-scripts/src/cli.ts
@@ -92,6 +92,7 @@ function getAllFiles(dirPath: string, arrayOfFiles: string[] = []) {
 function run() {
   const help = `
   Usage:
+    $ modular check
     $ modular add <package-name>
     $ modular start
     $ modular build
@@ -118,6 +119,8 @@ function run() {
         return start(argv._[1]);
       case 'build':
         return build(argv._[1]);
+      case 'check':
+        return check();
       default:
         console.log(help);
         process.exit(1);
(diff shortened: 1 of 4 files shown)


packages/modular-scripts/src/cli.ts
```diff
@@ -304,7 +307,20 @@ function build(appPath: string) {
...
+function check() {
+  const verifyPackageTree = require('react-scripts/scripts/utils/verifyPackageTree') as VerifyPackageTree; // eslint-disable-line @typescript-eslint/no-var-requires
+  verifyPackageTree();
+  // Unlike CRA, let's NOT verify typescript config when running tests.
+  console.log(chalk.greenBright(`All Good!`));
+}
+
 try {
+  if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
+    check();
```
threepointone, line 321: Calling this here will mean that every run of modular-scripts will start with "All Good!", but I think we want that only when check is called manually (especially because it'll then confuse someone of the following command fails). 

Also, calling modular check itself will output "All good!" twice. 

Further, the other scripts (build, start) calls this already within react-scripts. So if there's a failure it'll appear twice. So I don't think it's a good idea to call this by default for every script.

Maybe we shouldn't implement modular check right now (or rather, maybe verifyPackageTree() isn't the right implementation for it), or rather shouldn't be in this PR. Let's instead just solve the problem of errors being swallowed. Just call verifyPackageTree() in this process for test()?

threepointone (changes requested):

packages/modular-scripts/src/cli.ts
LukeSheard: > Also, calling modular check itself will output "All good!" twice.

lol I was so confused why this was happening - didn't think about it.

> Further, the other scripts (build, start) calls this already within react-scripts. So if there's a failure it'll appear twice. So I don't think it's a good idea to call this by default for every script. 

This is a good point - but verify will call `process.exit` if there is an issue, but I agree calling it twice is not ideal. 

> Just call verifyPackageTree() in this process for test()?

Yeah - I think for now this is the best implementation.

threepointone (approved):

PR #193, opened 2020-12-11
WIP; DO NOT MERGE: modular build <library>

threepointone:
This PR adds the feature to let `modular` build workspaces as libraries (as opposed to apps, which we currently support). It uses rollup to generate commonjs and esm builds of the workspace it's pointing to (based on the `main` field in the workspace's `package.json`), and then generates typescript typings for the same. It then generates a publishable folder under `dist/<workspace name>`, with a modified `package.json` that points to these built versions of the files. The folder will also contain any other files specified by the `files` field, and/or `.npmignore` in the workspace.

This is a port of the script we used internally at JPM for the same, and makes it a first class feature for modular. It's fast and works fairly well.

There are some major TODOs to do before this PR lands. 

- [ ] There are no tests! I'll add some. 
- [ ] This addresses doing a single build for a package (i.e. `modular build <name>`), but not for multiple packages at once (i.e. `modular build <name1>,<name2>,<name3>`), nor all packages at once (i.e. `modular build --all`).
- [ ] We should probably run whatever is in the workspace `package.json`'s `build` field too. 
- [ ] There are a number of TODOs specified in the build file itself (grep for `TODO:`). They're not all critical, but I'll try to do at least some of them. 
- [ ] We should use this to build `modular` packages themselves

Opening up this draft PR so I can get eyes on it, and to see whether the code is understandable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,8 @@
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
     "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
+  },
+  "resolutions": {
+    "rollup": "^2.33.3"
   }
 }
--- a/packages/modular-scripts/src/build.ts
+++ b/packages/modular-scripts/src/build.ts
@@ -1,0 +1,500 @@
+// While working on this file, be aware that builds
+// could be happening simultaneously across packages, so
+// try be 'thread-safe'. No state outside of functions
+
+import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
+import { JSONSchemaForTheTypeScriptCompilerSConfigurationFile as TSConfig } from '@schemastore/tsconfig';
+
+import { promisify as prom } from 'util';
+
+import * as rollup from 'rollup';
+import rimraf from 'rimraf';
+import * as path from 'path';
+import { extract } from 'tar';
+
+import execa from 'execa';
+
+import postcss from 'rollup-plugin-postcss';
+import json from '@rollup/plugin-json';
+import resolve from '@rollup/plugin-node-resolve';
+import babel from '@rollup/plugin-babel';
+import commonjs from '@rollup/plugin-commonjs';
+
+import * as ts from 'typescript';
+import * as fse from 'fs-extra';
+
+import builtinModules from 'builtin-modules';
+
+// from https://github.com/Microsoft/TypeScript/issues/6387
+// a helper to output a readable message from a ts diagnostics object
+function reportTSDiagnostics(diagnostics: ts.Diagnostic[]): void {
+  diagnostics.forEach((diagnostic) => {
+    let message = 'Error';
+    if (diagnostic.file) {
+      const where = diagnostic.file.getLineAndCharacterOfPosition(
+        diagnostic.start as number,
+      );
+      message += ` ${diagnostic.file.fileName} ${where.line}, ${
+        where.character + 1
+      }`;
+    }
+    message +=
+      ': ' + ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n');
+    console.log(message);
+  });
+}
+
+const extensions = ['.js', '.jsx', '.ts', '.tsx'];
+const outputDirectory = 'dist';
+const typescriptConfigFilename = 'tsconfig.json';
+const packagesRoot = 'packages';
+
+// list of all directories under packages
+const packageDirectoryNames = fse
+  .readdirSync(packagesRoot, { withFileTypes: true })
+  .filter((directoryEntry) => directoryEntry.isDirectory())
+  .map((directory) => directory.name);
+
+// dependencies defined at the root
+const rootPackageJsonDependencies =
+  (fse.readJSONSync('package.json') as PackageJson).dependencies || {};
+
+// a map of all package.json contents, indexed by package name
+const packageJsons: { [key: string]: PackageJson } = {};
+// a map of all package.json contents, indexed by directory name
+const packageJsonsByDirectoryName: {
+  [key: string]: PackageJson;
+} = {};
+
+// let's populate the above two
+for (let i = 0; i < packageDirectoryNames.length; i++) {
+  const pathToPackageJson = path.join(
+    packagesRoot,
+    packageDirectoryNames[i],
+    'package.json',
+  );
+  if (fse.existsSync(pathToPackageJson)) {
+    const packageJson = fse.readJsonSync(pathToPackageJson) as PackageJson;
+    // TODO: throw if there isn't a name?
+    packageJsons[packageJson.name as string] = packageJson;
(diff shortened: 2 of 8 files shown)


packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,488 @@
+// While working on this file, be aware that builds
```
threepointone, line 1: annoying that github is hiding this file because it's 'too big'.

tsconfig.json
```diff
@@ -20,6 +20,7 @@
 
     /* Advanced Options */
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
-    "sourceMap": true
+    "sourceMap": true,
+    "incremental": true
```
threepointone, line 24: used this opportunity to enable incremental typechecking for our repository, but we should bring it to every modular repository by default

package.json
```diff
@@ -64,5 +64,8 @@
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
     "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
+  },
+  "resolutions": {
```
threepointone, line 68: super annoying that I had to do this, but a deep dependency inside react-scripts brings an older version of rollup and ruins types without this line. thinking of a better fix.

packages/modular-scripts/src/build.ts
threepointone, line 1: I'm looking for feedback on whether the comments in this file are easy to understand, and if the code itself is grokkable.

LukeSheard: Could we break this up to make to it easier to grok / review in GitHub then?

threepointone: It's 500 lines of commented procedural code grouped by function, it's not that much. This is on github to fix.

threepointone: Well, maybe I could split it up. Ok, I promise to think strongly about it.

packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,488 @@
...
+// could be happening simultaneously across packages, so
+// try be 'thread-safe'. No state outside of functions
+
+import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
+import { JSONSchemaForTheTypeScriptCompilerSConfigurationFile as TSConfig } from '@schemastore/tsconfig';
+
+import { promisify as prom } from 'util';
+
+import * as rollup from 'rollup';
```
NMinhNguyen, line 10: I wish that we'd had a chance to discuss this as a team prior to an implementation being written, but there's a big problem with this bundling approach for libraries. You can read more about it at https://webpack.js.org/blog/2020-12-08-roadmap-2021/#esm-library and https://material-ui.com/guides/minimizing-bundle-size/

threepointone: The first link is a 404, and it's not clear what you're suggesting in link 2. 

I've brought this up multiple times inside the team over the last couple of months, pointed to the internal PR, gave constant updates, did a walkthrough of the code, and asked for people to look at it multiple times. 

Please give specific feedback and I can address it.

NMinhNguyen: The first link is a 404 because of Service Workers, clear your cache/do a hard reset. The specific feedback is explained in detail if you navigate to the first link, but the idea is it's harmful for modular (regular use of the word modular, not Modular) libraries to be shipped as a bundle because of tree shaking issues.

threepointone: Noted as a possible tradeoff, though it would be weird if a bundler included functions in an es module that weren't referenced (I'll verify the behaviour later). imo not a deal breaker for libraries that aren't large collections of utilities (and if so, they don't have to `modular build` to build those workspaces).

NMinhNguyen: And the reason I brought up that Material-UI link is because they intentionally do not distribute their package as an ESM/CJS bundle, but rather they ship unbundled modules so that tree-shaking can work reliably.

tsconfig.json
NMinhNguyen: Check https://github.com/vercel/next.js/issues/8708#issuecomment-570384682 for why `incremental` doesn't make sense in the context of a project that uses Babel to transpile TypeScript.

packages/modular-scripts/src/build.ts
threepointone: I'll look at preserving module structure, shouldn't be too hard.

tsconfig.json
threepointone: It's still useful while running typechecking locally (and can confirm `yarn tsc` is noticeably faster)

packages/modular-scripts/src/build.ts
NMinhNguyen: Going back to my original comment, I strongly suggest that you revisit the question of whether it even makes sense to bundle a library (webpack says it doesn't). Webpack are aware of the limitation that it can't produce ESM bundles, but they say that it's actually not something you'd want in the first place, hence why it isn't being prioritised.

threepointone: Yes, that's why I'm looking at preserving the module structure.

tsconfig.json
NMinhNguyen: Perhaps if it's solely for typechecking, then `--incremental` can be used instead? https://github.com/vercel/next.js/issues/14997#issuecomment-726692314 Mostly depends on what else uses `tsconfig.json` and whether the presence of `incremental` can get in the way of that.

packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,488 @@
...
+        mainFields: ['module', 'main', 'browser'],
+      }),
+      commonjs({ include: /\/node_modules\// }),
+      babel({
+        babelHelpers: 'bundled',
+        presets: [
+          ['@babel/preset-typescript', { isTSX: true, allExtensions: true }],
+          '@babel/preset-react',
+          [
+            '@babel/preset-env',
```
NMinhNguyen, line 182: Another important consideration here is overbundling: https://web.dev/publish-modern-javascript/ and https://next.material-ui.com/guides/minimizing-bundle-size/#modern-bundle when apps inside the monorepo are built, are they going to use output produced by this build when they point at packages in the monorepo?

threepointone: No, they use the source directly.

NMinhNguyen: That's good. Although even for published packages we should perhaps set a lower bound and not transpile below IE11, for example.

packages/modular-scripts/src/build.ts
threepointone: This is updated to now preserve the module/directory structure, and doesn't make bundles.

threepointone:
Closing this PR while I work on the todos.

PR #194, opened 2020-12-11
Initial prototype of storybook integration

LukeSheard:
Introduce a storybook command (right now using storybook itself, the idea being that it would be and agnostic command if we move to something else). 

Picks up all `*.stories.(js|ts|tsx)` files in packages. 

- [ ] Do we restrict to `packages` only? Are apps supposed to have stories?

### DIFF
--- a/packages/create-modular-react-app/template/gitignore
+++ b/packages/create-modular-react-app/template/gitignore
@@ -8,6 +8,9 @@ node_modules
 # testing
 /coverage
 
+# storybook
+/storybook-static
+
 # production
 /packages/*/build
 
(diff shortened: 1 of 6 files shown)


threepointone:
This is starting to look good. Based on our conversation, please figure out how we can run this ourselves without building the workspace, a sample .stories.js file to mess with, and some comments on the code. 

(For posterity) I was thinking we should also have a conversation whether storybook should actually be included ootb with modular (or possibly installed dynamically on usage?) but eh... it's useful enough that I can't think of why we *shouldn't* have this if it means better UI development. 

Thanks for the PR! Looking forward to this now.

threepointone:
The tests hung on CI and took up 6 hours of our minutes btw. 

I'm also feeling more conservative now on this, and I'm not sure why we should have storybook integration ootb. Doesn't seem fundamental.

threepointone:
As discussed, maybe this shouldn't belong in modular core, but we must definitely provide first class support inside JPM.

PR #195, opened 2020-12-18
`modular build <library>`

threepointone:
This PR adds a feature to let modular build workspaces as libraries (as opposed to only apps). It uses rollup to generate commonjs and esm builds of the workspace it's pointing to (based on the `main` field in the workspace's `package.json`), and then generates typescript typings for the same. It then generates a publishable folder under `dist/<workspace name>`, with a modified `package.json` that points to these built versions of the files. The folder will also contain any other files specified by the `files` field, and/or `.npmignore` in the workspace. Passing `--preserve-modules` decides whether or not to make a bundle. 

You can also do parallel builds by passing a comma-separated list of packages: `modular build app1,app2,package1,view1,etc`

Finally, I made it so app builds go to `/dist/<app-name>`, for consistency. 

This is a port of the script we used internally at JPM for the same, and makes it a first class feature for modular. It's fast and works fairly well.

Three things missing from this:

- We should use it for our own builds 
- No tests for parallel/multiple builds
- Docs

I'll do these in the next PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,8 @@
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
     "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
+  },
+  "resolutions": {
+    "rollup": "^2.33.3"
   }
 }
--- a/packages/modular-scripts/src/__tests__/index.test.ts
+++ b/packages/modular-scripts/src/__tests__/index.test.ts
@@ -1,6 +1,5 @@
 import execa from 'execa';
 import rimraf from 'rimraf';
-import puppeteer from 'puppeteer';
 import tree from 'tree-view-for-tests';
 import path from 'path';
 import fs from 'fs-extra';
@@ -27,17 +26,6 @@ const { getNodeText } = queries;
 jest.setTimeout(10 * 60 * 1000);
 
 const packagesPath = path.join(getModularRoot(), 'packages');
-
-function getBrowser() {
-  return puppeteer.launch(
-    process.env.CI
-      ? {
-          headless: true,
-          args: ['--no-sandbox', '--disable-setuid-sandbox'],
-        }
-      : {},
-  );
-}
 
 function modular(str: string, opts: Record<string, unknown> = {}) {
   return execa('yarnpkg', ['modular', ...str.split(' ')], {
@@ -190,6 +178,28 @@ describe('modular-scripts', () => {
   });
 
   it('can start an app', async () => {
+    // Ok, so. Sunil's decided to get the new M1 MacBook Air. Some software doesn't run on it
+    // well yet. Particularly the puppeteer npm package failes to install and run
+    // (see https://github.com/puppeteer/puppeteer/issues/, issues #6634 and #6641,
+    // possible fix in pull #6495)
+
+    // Because of this, he's marked puppeteer in optionalDependencies, so it's failure to install
+    // doesn't block everything else. Further, because this particular test is already flaky,
+    // it's disabled when running locally. However, because it fails to install, it causes
+    // typescript and eslint failures. Hence the need to disable those errors for now.
+
+    // It's Sunil's responsibility to fix this when better, so shout at him if he doesn't.
+
+    /* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call */
+
+    // This seems to be leaving hanging processes locally,
+    // so marking this test as a no-op for now. Sigh.
+    if (!process.env.CI) {
+      return;
+    }
+
+    const puppeteer = require('puppeteer');
+    // @ts-expect-error FIXME
     let browser: puppeteer.Browser | undefined;
     let devServer: DevServer | undefined;
     try {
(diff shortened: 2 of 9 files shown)


packages/modular-scripts/src/cli.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from 'change-case';
 import prompts from 'prompts';
 import resolveAsBin from 'resolve-as-bin';
-
+import buildPackage from './build';
```
LukeSheard, line 15: Build does a lot of top level work - so potentially we want to hide this import or change the top level work done?

packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,619 @@
...
+    ],
+    // TODO: support for css modules, sass, dotenv,
+    // and anything else create-react-app supports
+    // (alternatively, disable support for those in apps)
+  });
+
+  const outputOptions = {
+    freeze: false,
+    sourcemap: true, // TODO: read this off env
+    globals: { react: 'React', 'react-native': 'ReactNative' }, // why?
```
LukeSheard, line 309: Agreed - why?

package.json
```diff
@@ -64,5 +64,8 @@
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint --cache --fix --ext .js,.ts,.tsx --max-warnings 0",
     "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
+  },
+  "resolutions": {
+    "rollup": "^2.33.3"
```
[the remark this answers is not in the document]
threepointone: Good catch, I'd meant to fix this.

packages/modular-scripts/src/build.ts
threepointone: will just remove this, doesn't make sense for us.

packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,619 @@
...
+        where.character + 1
+      }`;
+    }
+    message +=
+      ': ' + ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n');
+    console.log(message);
+  });
+}
+
+const extensions = ['.js', '.jsx', '.ts', '.tsx'];
```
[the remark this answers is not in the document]
threepointone: I don't think this is a prefer ranking? Thought it was just a set of extensions we recognise (for the plugins it's used below).

packages/modular-scripts/src/cli.ts
threepointone: I didn't understand this comment, could you clarify what you mean? Do you want me to move build() into this file?

packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -189,6 +178,28 @@ describe('modular-scripts', () => {
   });
 
   it('can start an app', async () => {
```
threepointone, line 180: Sigh.

packages/modular-scripts/src/build.ts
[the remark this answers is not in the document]
LukeSheard: I was just double checking - because in webpack the extensions property is an ordered list. I assumed that this would be the same for roll up.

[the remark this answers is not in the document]
threepointone: I'll flip it around anyway.

packages/modular-scripts/src/__tests__/index.test.ts
LukeSheard: Oh the problems of buying a new MacBook. 😄

packages/modular-scripts/src/cli.ts
LukeSheard: My main concern is that there are import side-effects for `build.ts` e.g. [1] / [2] which do a lot of FS operations. On a machine with a slow-FS (e.g. network drives) this could slow down the startup of the cli for all operations. I think we should either:

1. Remove the globals from the `build.ts` script and have them as locals for the build operations, or. 
2. Lazily import `build.ts` so that the initialisation only happens when it's needed. 

[1] https://github.com/jpmorganchase/modular/pull/195/files#diff-a3c185d299c664d2bc8bbc2bbade48ca2a96a5ed03c585d268e5290e97016c0aR77
[2] https://github.com/jpmorganchase/modular/pull/195/files#diff-a3c185d299c664d2bc8bbc2bbade48ca2a96a5ed03c585d268e5290e97016c0aR96

threepointone: Good points. Ok lemme rewrite this.

package.json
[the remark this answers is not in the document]
LukeSheard: Ah I've wished for a long time that we could have comments in `package.json` (yay for `tsconfig.json` syntax.)

As you said if this is fine then we can resolve this - I just wanted to raise the point to make sure I understood.

packages/modular-scripts/src/build.ts
```diff
@@ -0,0 +1,618 @@
...
+}
+
+const typescriptConfig: TSConfig = {};
+// validate tsconfig
+{
+  // Extract configuration from config file and parse JSON,
+  // after removing comments. Just a fancier JSON.parse
+  const result = ts.parseConfigFileTextToJson(
+    typescriptConfigFilename,
+    fse.readFileSync(typescriptConfigFilename, 'utf8').toString(),
```
LukeSheard, line 132: Perhaps these reads should be prefixed with `getModularRoot`?

threepointone: also good catch, will do

threepointone: Ok, instead, I've made this script error if it's not run from modular root. Makes it more broadly applicable across this file.

packages/modular-scripts/src/cli.ts
threepointone: Ok this import is now deferred till the time it's actually called.

packages/modular-scripts/src/build.ts
LukeSheard: LGTM

LukeSheard (approved):

PR #198, opened 2020-12-22
put a scary warning on the craco config file name

threepointone:
I noticed that folks reach directly to the config and use craco to run their apps. That sucks, and will make our upgrade path harder. Instead, we should work with them and either fix our config, or propose workarounds. This PR puts a scary disclaimer in the file name. In the future we'll remove craco altogether.

### DIFF
--- a/.changeset/tame-spiders-grin.md
+++ b/.changeset/tame-spiders-grin.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': patch
+---
+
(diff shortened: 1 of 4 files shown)


LukeSheard (approved):
Seems fair enough - I think right now while we fill out features we know there are some internal projects which are hacking around on this. But I think this is a well intended change. Also easy to grep for internally.

Issue #68, opened 2020-09-01
add .editorconfig to generated repos


threepointone:
If this is about owning the toolchain, we have the opportunity of setting up some default extensions and other dx-y stuff for freshly created repositories. (e.g., the eslint extension isn't installed by default with vscode, but seems fairly critical.) What would we add to this?

PR #199, opened 2020-12-22
Some basic settings for vscode

threepointone:
This is the start of the work to automate project settings in IDEs. This PR makes it so that every modular project, when opened in vscode, gets a recommendation to install the eslint plugin, some editor settings, and a pointer to typescript so it picks up the right version. We'll expand on this in the future.

Closes https://github.com/jpmorganchase/modular/issues/68, but we'll do more in the future. 

Additionally: while working on this, I discovered and fixed a bug where we weren't copying over all files from the template when creating a new project.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["dbaeumer.vscode-eslint"]
+}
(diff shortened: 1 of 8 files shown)


LukeSheard:
LGTM - progress is progress.

threepointone:
dammit, should've added a changeset to this.

Issue #110, opened 2020-09-23
What does the modular upgrade experience look and feel like?


threepointone:
This is a main issue to discuss what the process of updating versions of modular in a project should be like. While I don't expect any _specific_ outcomes in this issue, and it's likely we'll have more targeted solutions per release, I think it would be useful to talk about what our philosophy around them should be, and maybe discuss examples. Some of my early notes - 
- We should strive to have as few breaking changes as possible. While this might imply that we may suffer from sunk cost fallacy for design decisions, instead I suggest it means we be very intentional about the design decisions that we _do_ take. So, introducing new primitives or APIs should be scrutinised deeply. Preferring convention over configuration, and then enforcing that convention with a tool, is preferred, because if and when we _do_ make a change, it'll be possible to document specific steps to move away from it, and/or automate the change. 
- On that note, codemods. As an example, when we added multiple apps and rearranged what the folder structure was like, we could have written a codemod that updated the repository structure (rather, a 'filemod'). We didn't need to because not many people were using modular so we could do it by hand, but it was always an option. 
- Similarly, we should try to avoid patterns that will be hard to move away from (or replace it with a pattern that we can own). As an example https://github.com/jpmorganchase/modular/issues/94 (I'm not fully sold on this myself just yet, but it's a good example for this point)
- While we'll have changelogs, we might need a blog, or some docs experience that's better than "go to the /docs folder on github" that shows how to update manually (if at all) and if there are any changes etc.

PR #203, opened 2020-12-23
auto-update flow

LukeSheard:
Add a prompt during startup to validate whether users wish to update their version of modular when there is a new one available in the registry their repository is aligned with. Added wrapper around fetch to deal with HTTP / HTTPS proxies.

Screenshot 2020-12-23 at 6 54 41 pm


packages/modular-scripts/src/cli.ts
```diff
@@ -59,9 +62,7 @@ function isYarnInstalled(): boolean {
 
 type PackageType = 'app' | 'view' | 'root'; // | 'package', the default
 
-export interface PackageJson {
-  name: string;
-  private?: boolean;
+export type ModularPackageJson = PackageJson & {
```
threepointone, line 65: I can't recollect why this was previously exported, and I can't find an import for this type elsewhere anywhere now. Remove this export?

packages/modular-scripts/src/fetch.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (process.env.HTTP_PROXY || process.env.HTTPS_PROXY) {
+    return proxyAgent;
+  }
+};
+const proxyAgent = (parsedUrl: URL) => {
+  const protocol = parsedUrl.protocol;
+  if (protocol === 'https') {
+    return new HttpsProxyAgent(process.env.HTTPS_PROXY as string);
+  } else {
+    return new HttpsProxyAgent(process.env.HTTP_PROXY as string);
```
threepointone, line 14: sorry for my noob question, but will this work with work on http? There appears to be an `http-proxy-agent` module, should you be using that?

packages/modular-scripts/src/cli.ts
```diff
@@ -343,6 +339,44 @@ async function build(directoryName: string, preserveModules?: boolean) {
...
+    const newVersion = modualarScriptsRegistry["dist-tags"]["latest"];
+
+    const { version }: ModularPackageJson = fs.readJSONSync(path.join(__dirname, "..", "package.json"));
+    if (newVersion != version) {
+      const { dependencies }: ModularPackageJson = fs.readJSONSync(path.join(getModularRoot(), "package.json"));
+
+      const update  = await prompts({
+        type: 'confirm',
+        name: 'value',
+        message: `Modular is out of date - would you like to update from ${version} to ${newVersion}`,
```
threepointone, line 363: `modular` is lowercase. Also a suggestion for the wording. 
```suggestion
        message: `Your version of modular is ${version} . Would you like to update to ${newVersion}?`,
```

packages/modular-scripts/src/cli.ts
```diff
@@ -343,6 +339,44 @@ async function build(directoryName: string, preserveModules?: boolean) {
...
+  if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
+    const { stdout: registry } = await execa('yarnpkg', ['config', 'get', 'registry']);
+
+    const url = String(new URL("/modular-scripts", registry));
+
+    const res = await fetch(url);
+    const modualarScriptsRegistry: NPMRegistryListing = await res.json();
+    const newVersion = modualarScriptsRegistry["dist-tags"]["latest"];
+
+    const { version }: ModularPackageJson = fs.readJSONSync(path.join(__dirname, "..", "package.json"));
```
threepointone, line 356: This doesn't seem right, this will only work in our own repository (which is... odd, what does updating mean then). Did you mean to use dependencies['modular-scripts'] and compare with `semver` instead?

Also, instead of using package.json, maybe you should be using yarn.lock instead?

packages/modular-scripts/src/cli.ts
```diff
@@ -343,6 +339,44 @@ async function build(directoryName: string, preserveModules?: boolean) {
...
+
+async function preflight() {
+  if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
+    const { stdout: registry } = await execa('yarnpkg', ['config', 'get', 'registry']);
+
+    const url = String(new URL("/modular-scripts", registry));
+
+    const res = await fetch(url);
+    const modualarScriptsRegistry: NPMRegistryListing = await res.json();
+    const newVersion = modualarScriptsRegistry["dist-tags"]["latest"];
```
threepointone, line 354:
```suggestion
    const modularScriptsRegistry: NPMRegistryListing = await res.json();
    const newVersion = modularScriptsRegistry["dist-tags"]["latest"];
```

threepointone (commented):
A problem here is that it'll bother people to update every time they run, and that could be annoying. Also you'd have to disable this command on CI. Instead, how about just logging the message to update (with a copy paste command `yarn upgrade modular-scripts eslint-config-modular-app`, and let them do it whenever they want?

packages/modular-scripts/src/cli.ts
threepointone: Oh it'll look at the package.json of itself. I think you should use find-up to find the nearest package.json instead, because you can't guarantee that this file will be at this path (especially after we migrate to using `modular build` to build modular-scripts)

threepointone:
This is super impactful, wish we'd done it from the beginning. Thanks for the PR on your time off!
Trying to think of how we can write a tests for this 🤔

packages/modular-scripts/src/fetch.ts
LukeSheard: lol this is my noob mistake - I wish there was an abstracted version of this as a package. I will just take this for inspiration... https://github.com/touv/node-fetch-with-proxy/blob/master/src/index.js#L53

packages/modular-scripts/src/cli.ts
LukeSheard: 👍

packages/modular-scripts/src/cli.ts
LukeSheard: If I can work out jest mocks then it should be easy enough to mock out the `fetch` and the `fse` call to return differing things, then just assert that prompts was called?

LukeSheard:
> Trying to think of how we can write a tests for this 🤔

If I can work out jest mocks then it should be easy enough to mock out the `fetch` and the `fse` call to return differing things, then just assert that prompts was called?

packages/modular-scripts/src/cli.ts
threepointone: Mocking fetch is a dope idea. Yeah I think you're right, that should do it.

threepointone:
Mocking fetch is a dope idea. Yeah I think you're right, that should do it.

packages/modular-scripts/src/cli.ts
LukeSheard: I went around in circles a couple of times around this. There's also the question of how we deal with a major releases (once we get to `1.0.0`) and make that transition easy for people - although this preflight would give us the option to run codemods on install as well automatically.

LukeSheard:
> A problem here is that it'll bother people to update every time they run, and that could be annoying.

I think the easiest thing to do on this front is to add a `last check time` and only run once a (day|week) so that we don't spam too hard. Although I think we should push people to be on the latest version as much as possible as upgrading should be minimal effort and they get dope features. 💯

NMinhNguyen:
Is there any reason not to use a package like https://github.com/yeoman/update-notifier instead? It's what `npm` CLI uses.

threepointone:
>  Although I think we should push people to be on the latest version as much as possible as upgrading should be minimal effort and they get dope features. 💯

I agree in principle, but consider the workflow. They'll see this prompt while they're working on other stuff, and won't be able to update immediately. It'll likely happen in a separate PR where they can isolate this upgrade. Also, if we have breaking changes, it won't be just the package that has to be updated, they'll have to make some other changes too. For this to happen safely we'll require at the minimum a `modular check` command, if not some form of automated codemod/filemod process (related issue: https://github.com/jpmorganchase/modular/issues/110). So until that happens, I don't think it should be a prompt. 

In general we shouldn't deprecate versions if it's working fine (Imagine if yarn 1 kept asking us to update to yarn 2. I'd rather not use yarn at all then.). It's a different issue *inside* JP Morgan, and we'll pull different levers there to get teams to update (One thing to do is to literally update it ourselves haha)

To summarise, I don't think we should have the prompt _now_, but should *definitely* have it at some point when we have the full upgrade story in place. Fair?

tstordyallison:
Maybe it’s more of a single printed line saying it’s out of date + a hint to run some command to easy update? And it could show/check only once a day etc. 

Then it’s less in your face/blocking the run of the command you trying do right now. But a gentle prod to run the command when it’s a good time to.

threepointone:
Yup. To clarify, I meant we shouldn't have the "prompts()" call, but definitely have the message and copy-pastable command.

threepointone:
It can even show on every run then, that's not as big a problem.

packages/modular-scripts/src/preflight.ts
```diff
@@ -0,0 +1,52 @@
...
+async function isYarnInstalled(): Promise<boolean> {
+  try {
+    await execa('yarnpkg', ['-v']);
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
+
+async function preflight(): Promise<void> {
```
threepointone, line 17: could you call this (and the filename) `preflightCheck` instead? preflight is an adjective, not a good function name.

packages/modular-scripts/package.json
```diff
@@ -9,6 +9,7 @@
     "node": ">=12"
   },
   "scripts": {
+    "modular": "ts-node src/cli.ts",
```
threepointone, line 12: why?

yarn.lock
```diff
@@ -1765,6 +1765,11 @@
   resolved "https://registry.yarnpkg.com/@schemastore/tsconfig/-/tsconfig-0.0.9.tgz#77a1a1fb55eb879391aa78db3154dbcc4d93948b"
```
threepointone, line 1765: wow, this is a lot of dependencies. wild.

packages/modular-scripts/src/preflight.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
+    const { name, version } = fs.readJSONSync(
+      path.join(__dirname, '..', 'package.json'),
+    ) as PackageJson;
+
+    const message =
+      'Modular is out of date - you can update from ' +
```
threepointone, line 24: modular is lowercase

packages/modular-scripts/src/preflight.ts
```diff
@@ -0,0 +1,52 @@
...
+  try {
+    await execa('yarnpkg', ['-v']);
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
+
+async function preflight(): Promise<void> {
+  if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
```
threepointone, line 18: disable this when CI=true as well

threepointone:
got any ideas on how to pass the CI failure?

yarn.lock
LukeSheard: Yeah I know - the world of 1 liner modules...

LukeSheard:
I'm not sure how we fix the CI since the error is coming from the fact we use `isolatedModules` to compile files, which basically means you can then only use a subset of the TypeScript language. I couldn't find any open issues against it so I'm not sure if there's any workaround. This likely opens up the debate about what the *right* TypeScript compiler flags..

threepointone:
Hmm it's because we use babel-plugin-typescript, so isolatedModules is enforced. Lemme think about it.

LukeSheard:
> it's because we use babel-plugin-typescript, so isolatedModules is enforced

Yeah - this was one thing that always frustrated me about the babel TypeScript eco-system. 

I think the easiest issue might be to patch the typings themselves - maybe we can add some kind of patch on the node_modules to work around since it's not our code. We can't use the official babel workaround see [1] point 2. 

[1] https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats

package.json
```diff
@@ -66,6 +66,7 @@
     "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
   },
   "resolutions": {
-    "rollup": "^2.33.3"
+    "rollup": "^2.33.3",
+    "**/update-notifier/boxen": "^5.0.0"
```
threepointone, line 70: why not wait for your update-notifier PR to land?

packages/modular-scripts/package.json
LukeSheard: This is just useful if you're hanging out in the `modular-scripts` folder itself rather than needing two prompts for each place in the repo.

package.json
LukeSheard: I think the real fix we need is in the DefinitelyTyped PR which currently has CI failing - so this is a workaround for now to get them all in line.

threepointone (approved):

PR #206, opened 2020-12-25
Use eslint-config-modular-app for our own project

threepointone:
This PR uses eslint-config-modular-app for our own app. It also takes some of the custom rules we were using and adds them to the config.

This also adds .prettierignore files, for ourselves, and generated projects.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,108 +1,33 @@
-// The ESLint browser environment defines all browser globals as valid,
-// even though most people don't know some of them exist (e.g. `name` or `status`).
-// This is dangerous as it hides accidentally undefined variables.
-// We blacklist the globals that we deem potentially confusing.
-// To use them, explicitly reference them, e.g. `window.name` or `window.status`.
-// See https://github.com/facebook/create-react-app/blob/3c2f2d4/packages/eslint-config-react-app/index.js#L19-L24
-const restrictedGlobals = require('confusing-browser-globals');
+'use strict';
 
+// Short hand codes for the values eslint expects
+// Might feel like overkill, but is also effectively visually
+// when scanning through the config and looking for values.
+
+// eslint-disable-next-line  no-unused-vars
 const ERROR = 'error';
 const WARN = 'warn';
 const OFF = 'off';
 
 module.exports = {
-  root: true,
-  env: {
-    browser: true,
-    es6: true,
-    node: true,
-  },
-  extends: [
-    'eslint:recommended',
-    'plugin:import/recommended',
-    'plugin:react/recommended',
-    'plugin:react-hooks/recommended',
-    'prettier',
-    'prettier/react',
-  ],
-  globals: {
-    Atomics: 'readonly',
-    SharedArrayBuffer: 'readonly',
-  },
-  parser: 'babel-eslint',
-  parserOptions: {
-    ecmaFeatures: {
-      jsx: true,
-    },
-    ecmaVersion: 2020,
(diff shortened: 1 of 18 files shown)


.eslintrc.js
```diff
@@ -1,108 +1,29 @@
-// The ESLint browser environment defines all browser globals as valid,
-// even though most people don't know some of them exist (e.g. `name` or `status`).
-// This is dangerous as it hides accidentally undefined variables.
-// We blacklist the globals that we deem potentially confusing.
-// To use them, explicitly reference them, e.g. `window.name` or `window.status`.
-// See https://github.com/facebook/create-react-app/blob/3c2f2d4/packages/eslint-config-react-app/index.js#L19-L24
-const restrictedGlobals = require('confusing-browser-globals');
+'use strict';
 
+// eslint-disable-next-line  no-unused-vars
```
LukeSheard, line 3: What's up with these vars anyway? Can we add a comment so to why they are here?

packages/eslint-config-modular-app/index.js
```diff
@@ -1,5 +1,42 @@
 'use strict';
 
+// eslint-disable-next-line  no-unused-vars
```
LukeSheard, line 3: See other comment - why are these here?

packages/modular-scripts/types/app/.eslintignore
```diff
@@ -0,0 +1 @@
+/public
```
LukeSheard, line 1: Is it not possible to just ignore these as part of the root eslint config? I think the idea of having scattered files around the codebase is confusing - if we can we should keep configuration in the root of the project so that you can conceptually hold it in your head from one source of truth.

packages/modular-views.macro/src/__tests__/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
LukeSheard, line 1: Is it not possible to just ignore these as part of the root eslint config? I think the idea of having scattered files around the codebase is confusing - if we can we should keep configuration in the root of the project so that you can conceptually hold it in your head from one source of truth.

packages/modular-scripts/types/app/.prettierignore
```diff
@@ -0,0 +1 @@
+/public
```
LukeSheard, line 1: Is it not possible to just ignore these as part of the root config? I think the idea of having scattered files around the codebase is confusing - if we can we should keep configuration in the root of the project so that you can conceptually hold it in your head from one source of truth.

packages/modular-scripts/types/app/.eslintignore
threepointone: done

packages/modular-scripts/types/app/.prettierignore
threepointone: done

packages/modular-views.macro/src/__tests__/.eslintrc.js
threepointone: this one I think is worth the exception.

packages/eslint-config-modular-app/index.js
threepointone: added a comment

.eslintrc.js
threepointone: added a comment

threepointone:
Going to land this, it's looking good. We can iterate more on it ofc.

PR #207, opened 2020-12-27
enhancements/fixes to `modular build <package>`

threepointone:
I've been working on using `modular build <package>` to build modular's packages. There's more to do, but these changes should probably land regardless. The changes I had to make during this PR

- include README.md in published packages by default
- fix dependencies in modular-scripts (the build script correctly caught the missing ones!)
- exclude /dist when generating type definitions
- exclude local imports from missingDependencies generated when rollup makes facades.
- fix our own tsconfig
- make the output a little flatter (re: directory structure). This helps in avoiding issues when using __dirname/__filename in node packages. (The problem still exists  if preserveModules: false, but that can't be avoided much. But even in that case, it's still better when the source files are flat in src)

There are more TODOs that arose from this, included in a comment at the bottom of build.ts, and I'll keep working on them.

### DIFF
--- a/packages/modular-scripts/src/build.ts
+++ b/packages/modular-scripts/src/build.ts
@@ -2,6 +2,9 @@
 // could be happening simultaneously across packages, so
 // try be 'thread-safe'. No state outside of functions
 
+// shorthand for building every workspace, if you're ever debugging this flow
+// rm -rf dist && yarn modular build `ls -m1 packages | sed -e 'H;${x;s/\n/,/g;s/^,//;p;};d'`
+
 import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
 import { JSONSchemaForTheTypeScriptCompilerSConfigurationFile as TSConfig } from '@schemastore/tsconfig';
 
@@ -160,6 +163,9 @@ const typescriptConfig: TSConfig = {};
       '**/*.spec.tsx',
       '**/*.test.tsx',
       '__tests__',
+      '**/dist-cjs',
+      '**/dist-es',
+      '**/dist-types',
       // TS defaults below
       'node_modules',
       'bower_components',
@@ -250,6 +256,7 @@ async function makeBundle(
     input: path.join(packagesRoot, directoryName, packageJson.main),
     external: (id) => {
       // via tsdx
+      // TODO: this should probably be included into deps instead
       if (id === 'babel-plugin-transform-async-to-promises/helpers') {
         // we want to inline these helpers
         return false;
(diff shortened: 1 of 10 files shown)


packages/modular-scripts/src/build.ts
```diff
@@ -250,6 +256,7 @@ async function makeBundle(
     input: path.join(packagesRoot, directoryName, packageJson.main),
     external: (id) => {
       // via tsdx
+      // TODO: this should probably be included into deps instead
```
LukeSheard, line 259: I think we should start up for a strategy for managing all the TODOs - either a central file we can view them all or extract them into issues (or both linked?)

packages/modular-scripts/src/build.ts
```diff
@@ -465,28 +472,35 @@ async function makeBundle(
     // TODO: what of 'bin' fields?
     main: preserveModules
```
LukeSheard, line 473: What’s the flow for using preserve modules? What’s the right way for consumers to use this flag? I would imagine that it would actually make more sense as a flag in the modular metadata in package.json for the package so that this flag is abstracted from the commands, otherwise I foresee people using it differently within a team when working.

packages/modular-scripts/src/build.ts
```diff
@@ -559,8 +573,16 @@ export async function build(
   // ensure the root build folder is ready
   await fse.mkdirp(outputDirectory);
 
-  // delete any existing build folder, if at all
-  await prom(rimraf)(path.join(packagesRoot, directoryName, outputDirectory));
+  // delete any existing local build folders
```
LukeSheard, line 576: What do we do if one of these commands fails? People are left with half created modules? Maybe we should add some cleanup here so we don’t leave rouge files around.

packages/modular-scripts/src/getModularRoot.ts
```diff
@@ -1,4 +1,3 @@
-// keep this in sync with modular-views.macro/src/getModularRoot
 import * as path from 'path';
```
LukeSheard, line 1: I feel like we should just expose this API as we’ll end up needing it in multiple places anyway - or it should just be a stand-alone that we can share.

packages/modular-scripts/src/build.ts
threepointone: These are non critical TODOs so I've left them inline. Happy to organise them later.

packages/modular-scripts/src/build.ts
threepointone: Yeah if you see the list of TODOs at the bottom of this file, having it as metadata in package.json is a line item. Considering this an unreleased feature for now so I haven't documented a flow yet. Want to use it at JPM and our own repo and then make a call whether it makes sense.

packages/modular-scripts/src/build.ts
threepointone: Also one of the followup items in the list at the bottom of this file.

packages/modular-scripts/src/getModularRoot.ts
threepointone: I think It's fine to duplicate this code, particularly because it's used differently at different spots, so they're not actually in sync.

threepointone (commented):
replied.

LukeSheard (approved):

PR #210, opened 2020-12-29
Fix build.ts type error

LukeSheard:
Fixes the issue with `any` assignment in `build.ts`

### DIFF
--- a/packages/modular-scripts/src/build.ts
+++ b/packages/modular-scripts/src/build.ts
@@ -148,6 +148,8 @@ const typescriptConfig: TSConfig = {};
     throw new Error('Failed to load Typescript configuration');
   }
 
+  const exclude: string[] = (configObject.exclude as string[]) || [];
+
   Object.assign(typescriptConfig, configObject, {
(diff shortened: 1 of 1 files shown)


packages/modular-scripts/src/build.ts
```diff
@@ -148,6 +148,8 @@ const typescriptConfig: TSConfig = {};
     throw new Error('Failed to load Typescript configuration');
   }
 
+  const exclude: string[] = (configObject.exclude as string[]) || [];
```
threepointone, line 151: Could you add a comment here why this is necessary?

threepointone (commented):
Please add a patch changeset.

packages/modular-scripts/src/build.ts
LukeSheard: Done

LukeSheard:
I don't think this needs a changeset since there's nothing to release here as part of this change.

threepointone (approved):

PR #212, opened 2020-12-30
Restrict tsconfig files so that TypeScript files live in `src` for packages

LukeSheard:
This PR restricts the location of TypeScript files into the `src` directory for each package to prevent erroneous files being picked up by type checking and build scripts. This is the beginning of imposing sensible defaults for the modular environment.

After replacing the JPM build script that we had pre `modular build` I came across this bug when trying to generate typescript files. Since the include is set so broadly it will start picking up generated `dist` files and fail. I edited the file locally to test and it works fine.

### DIFF
--- a/.changeset/wet-nails-care.md
+++ b/.changeset/wet-nails-care.md
@@ -1,0 +1,6 @@
+---
+'modular-scripts': patch
+---
+
+Fix an issue where generated files from libraries would interfere with
+typescript definition generation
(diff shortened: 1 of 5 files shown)


threepointone:
The fix should be to exclude /dist folders, which I've done in my PR.

LukeSheard:
I think restricting the included files is a better fix - since it minimizes the files to filter upfront.

threepointone:
That's a different conversation to have then, because that implies also changing the tsconfig, ad saying we should only have .ts(x) files under src. Which we can totally consider, but isn't the right fix for this specific purpose.

threepointone:
Pass CI and I'll stamp this?

threepointone (approved):

PR #215, opened 2020-12-30
When building libs, don't throw an error on submodule import for non-js files

threepointone:
Doing a submodule import is fine for static assets like css and images _for now_, so we should allow them when building. We should only do this for imports from local workspaces, not for third party deps.

### DIFF
--- a/.changeset/little-maps-swim.md
+++ b/.changeset/little-maps-swim.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': minor
+---
+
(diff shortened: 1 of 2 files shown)


threepointone:
I'll land this as soon as CI passes to block internal usage. Sorry for the repeated pushes, this is actually tricky to get right. This current iteration should be fine (but not complete; it'll allow directory style imports). This should be revisited soon, but maybe it's better to enforce it with the lint rule instead of during build.

PR #217, opened 2020-12-30
ignore local build folders

threepointone:
In #207, I changed the build folder destinations, but forgot to change gitignore directives for the same. This PR adds them for us, and for newly generated projects.

This isn't critical since they get cleaned on build anyway, but good to have.


threepointone:
Landing as discussed

PR #219, opened 2020-12-31
use a central tsconfig

threepointone:
This PR moves `modular.tsconfig` into `modular-scripts/tsconfig.json`. It also makes our own project use it, and any new projects as well.


LukeSheard (approved):
We should make sure to deprecate / delete the NPM package once this is released.

Issue #202, opened 2020-12-23
Modular should use `modular build modular-scripts`


threepointone:
modular can now build libraries (https://github.com/jpmorganchase/modular/pull/195). We should use it to build modular itself.

PR #220, opened 2021-01-02
Migrate modular-views.macro to TypeScript

LukeSheard:
An initial cut of migrating the modular views macro to TypeScript. This brings the macro in line with the rest of the codebase so far.

I had an issue with getting the tests to use a TypeScript fixture file with babel - so I setup a prebuild hook into the tests for now. If someone who knows babel bettter than me can get it to work then be my guest.


threepointone:
want to rebase/merge this on master before a review? I think the tsconfigs might change, no?

threepointone (approved):
stamping, but I'd love to revisit this in the future to see if we can simplify stuff. Feels like a bunch of added complexity, and I'm not sure what's gotten better.

LukeSheard:
I think the next step here is completing #202 which will allow us to remove the extra `tsconfig` file and commands which were needed here.

PR #230, opened 2021-01-19
include @finos/perspective-webpack-plugin by default

threepointone:
This PR adds `@finos/perspective-webpack-plugin` to the webpack plugin for apps. This makes modular usable by a wider range of teams inside JPM, considering how many of them use perspective. This may be a temporary addition until we figure out how to have better builds from perspetive themselves directly. Further, this doesn't add it for rollup/library builds, so we're not going to advertiswe this too loudly just yet,

I added the sample perspective react app's code (https://github.com/finos/perspective/tree/master/examples/react) into our own sample app to verify that this works. It does!


texodus:
@threepointone If you're interested, starting in `0.6.1` we're aiming to make perspective practically configurable without a custom plugin (though `@finos/perspective-webpack-plugin` will continue to exist and will just add `worker-loader` and `file-loader`).  See https://github.com/finos/perspective/pull/1292

Perspective should default work without the plugin for all recent releases as well via its inline-build, which is the package's [`"module"` and `"browser"` fields](https://github.com/finos/perspective/blob/master/packages/perspective/package.json#L8);  in this build the WebAssembly asset is inlined as a base64-encoded string within the Webpack bundle.  This bloats the bundle a bit but requires no Webpack magic and has no other performance impact. See https://github.com/finos/perspective/pull/870

If you need any specific changes, questions, etc. re building Perspective - feel free to open a Perspective Issue or @ me, we are constantly seeking to improve it.

threepointone:
Thanks @texodus! I'll reach out to you soon.

packages/modular-site/src/App.tsx
```diff
@@ -1,3 +1,56 @@
...
+          void viewer.current.restore(config);
+        }
+      },
+      (err) => console.error(err),
+    );
+  }, []);
+
+  // You can also the use the stringified config values as attributes
+  return (
+    //eslint-disable-next-line react/react-in-jsx-scope
```
threepointone, line 51: we need to disable this in our eslint config too, this isn't needed anymore with the new jsx runtime

LukeSheard: Woah funky this is gonna change a lot!

LukeSheard:
This is pretty pragmatic - obviously if we could do this without a plugin that would be the best in the long term as this currently locks us into `webpack` for now though.

packages/modular-site/src/App.tsx
threepointone: Full on! Thankfully there's a codemod, that should be a fun exercise. createElement still works too, so it's backward-compatible.

LukeSheard (approved):

PR #232, opened 2021-01-21
Modular build (library) should fail loudly if there are setup issues

LukeSheard:
I just ran across a case where I'd setup library publishing - my `modular build` command was "working" locally but it wasn't until I went to publish in CI that I realised it was skipping a module which was setup incorrectly. When building multiple packages simultaneously the `bail` logging gets lost easily.  

I think that `modular build` for a library should fail loudly if there's an issue - a bit like webpack would fail loudly if we tried to build an app and there was an issue.

### DIFF
--- a/packages/modular-scripts/src/build.ts
+++ b/packages/modular-scripts/src/build.ts
@@ -199,59 +199,51 @@ async function makeBundle(
   const packageJson = packageJsonsByDirectoryName[directoryName];
 
   if (!packageJson) {
-    console.error(
+    throw new Error(
       `no package.json in ${packagesRoot}/${directoryName}, bailing...`,
     );
-    return false;
   }
   if (packageJson.private === true) {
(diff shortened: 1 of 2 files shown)


packages/modular-scripts/src/build.ts
```diff
@@ -199,59 +199,51 @@ async function makeBundle(
   const packageJson = packageJsonsByDirectoryName[directoryName];
 
   if (!packageJson) {
-    console.error(
+    throw new Error(
```
threepointone, line 202: I wasn't thinking this is an error, just a bypass. But I can see how people should be explicit in asking for the things to build, so yeah this should be an error.

threepointone (approved):

threepointone:
Add a changeset!

LukeSheard:
Changeset added 🚀

PR #235, opened 2021-01-22
Proposal: Enable multi-team-workspaces

LukeSheard:
Through testing internally I think it's become apparent that we need to be able to support different "verticals" within the same repository - this might be separate teams who don't share the same scope within their packages. 

We discussed this previously and said that we would have to be strict about the implementation - I've done a bit of a PoC for what this could look like. The rules work as follows: 
1. `workspaces` in the modular `root` `package.json` must either have `[ "packages/*"] ]` or `[ "packages/*/*" ]`.
2. In the single `*` namespace end. 
3. In the multi scope case...
    *  validate that each package name in the "scope" directory name matches.
    * validate that each package name matches the directory name which contains it.

This will enable the repo to scale across multiple teams and other `modular` based teams to merge repositories without changing their scope. It also means that `CODEOWNERS` can work more easily, since you can assign `packages/team/*` to a particular group.

### DIFF
--- a/packages/modular-scripts/src/preflightCheck.ts
+++ b/packages/modular-scripts/src/preflightCheck.ts
@@ -1,10 +1,11 @@
-import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
 import chalk from 'chalk';
 import execa from 'execa';
 import * as fs from 'fs-extra';
 import isCI from 'is-ci';
 import path from 'path';
 import updateNotifier from 'update-notifier';
+import getModularRoot from './getModularRoot';
+import { ModularPackageJson } from './isModularType';
 
 async function isYarnInstalled(): Promise<boolean> {
   try {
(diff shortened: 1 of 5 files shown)


packages/modular-scripts/src/preflightCheck.ts
```diff
@@ -48,6 +50,75 @@ async function preflightCheck(): Promise<void> {
...
+      } else {
+        const workspace = workspaces[0];
+        if (![`packages/*`, `packages/*/*`].includes(workspace)) {
+          throw new Error(
+            'modular expects a single workspaces property with either `packages/*` or `packages/*/*`',
+          );
+        }
+
+        // if we're in multi scope mode then packages must adhere to their respective scope
+        if (workspace === `packages/*/*`) {
```
threepointone, line 73: enforce this as a separate script, not in modular core. 
reason: this adds a new constraint for using modular, and breaks backward compatibility. 
I think it's useful, but let's do it in modular v2 (which we can start working on very soon).

NMinhNguyen:
> Through testing internally I think it's become apparent that we need to be able to support different "verticals" within the same repository - this might be separate teams who don't share the same scope within their packages.

I had to read the code to understand what was meant by 'scope' because my initial thinking (just reading the PR description) was it's to do with [npm scopes]( https://docs.npmjs.com/cli/v6/using-npm/scope) which would make sense given this is in the context of packages. Because of this, I'd be very careful when introducing terminology that can be easily misinterpreted (because of existing conventions).

threepointone:
I made an RFC that proposes a looser approach https://github.com/jpmorganchase/modular/issues/240, and associated implications. 

I also started working on a branch that tackles the implementation, and have already gotten pretty far with it. I'd recommend closing this, I'm happy to take this on and hopefully ship it relatively soon.

threepointone:
Closing this for https://github.com/jpmorganchase/modular/pull/246, thank you for getting it here!

PR #239, opened 2021-01-29
[DO NOT MERGE] use babel-preset-react-app for lib builds

threepointone:
Keeps output consistent across apps and libs, across upgrades and changes.
(For the babel part, we still need to fix some missing features.)
This PR should also make lib bundles smaller, by avoiding bundling @babel/runtime helpers. 

I need to test this internally before merging it in, so don't merge this just yet.

### DIFF
--- a/.changeset/gold-monkeys-leave.md
+++ b/.changeset/gold-monkeys-leave.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': patch
+---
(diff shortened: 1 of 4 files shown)


threepointone:
The breaking build is annoying, not reproducing locally, but I'mm punting on it till I verify the functionality internally anyway.

PR #242, opened 2021-01-30
fix the macro's type generation, and some dependency fixes

threepointone:
- `modular-views.macro` was generating to a different folder than what its package.json was pointing to, fixed that.
- added `@babel/core` as a dependency to `modular-scripts`, to avoid a whole bunch of missing peer dependency warnings whenever someone makes a new project.
- `@manypkg/cli check` also suggested a couple of changes for the root package.json that seemed reasonable, added those too.


threepointone:
gonna land this, it's annoying to keep bumping into the stray generated macro type.

PR #243, opened 2021-01-30
fix more dependencies / console spam.

threepointone:
- `@testing-library/dom` is a peerdependency for `@testing-library/user-event`, we should install it too in new projects.
- Installed both those packages locally for ourselves as well, just to keep us honest.
- we should install `eslint` in new projects too, much like we install `prettier`. It's already being installed so it doesn't add more to the load, but it does prevent a bunch of console spam.


threepointone:
gonna land this too, want to reduce the ick factor when running tests and such.

PR #245, opened 2021-01-30
use yarn --silent 

threepointone:
This adds the `--silent` flag when adding dependencies in a new project. It reduces a whole bund of console spam, but stills shows progress bars, and importantly, also stderr.

To compare, on my laptop, running all tests now is about 400 lines of console spam, compared to about 2400 lines before this PR. (Your numbers may change depending on terminal width, etc)


threepointone:
landing this one too

PR #268, opened 2021-02-15
E2E-tests

Cono52:
'e2e' command to run apps that are set up to run cypress, sequentially.


packages/modular-scripts/src/cli.ts
```diff
@@ -353,6 +372,81 @@ async function buildSequential(
   }
 }
 
+// currently assumes cypress/e2e tests are not set up at all.
+async function initE2E(someArgs: string) {
```
Cono52, line 375: Things that would make this "setup" useful is quite foggy at the minute.
It might be useful to have a modular type 'package' in the case things are added to the monorepo that do not have types i.e. are simply dragged in.

tsconfig.json
```diff
@@ -5,5 +5,8 @@
     "packages/**/src",
     "packages/modular-scripts/types/",
     "packages/create-modular-react-app/template/"
-  ]
+  ],
+  "compilerOptions": {
+    "typeRoots": ["./node_modules/@types", "./packages/modular-scripts/@types"]
```
Cono52, line 10: not sure why, but currently, the the types i put in modular scripts aren't being resolved WIP

Cono52:
Branch is stale -> moved to https://github.com/jpmorganchase/modular/tree/cypress-integration

Cono52:
some stuff in branch I want to refer back to so leaving un-deleted for the moment

PR #270, opened 2021-02-23
[DO NOT MERGE] don't delete generated artifacts on build

threepointone:
There's a race condition when we run `modular build <path>` on several packages at the same time, in different processes. `makeTypings()`, despite being synchronous, can be interleaved across processes. So, it may have picked up a built package, where the `package.json` is pointing to built typings. When we delete the built folders, the generated types don't exist anymore (but it won't pick up the original source), making the build fail.

To fix this for now, we're just not going to delete the generated artifacts.

I also updated prettierignore/eslintignore/gitignore in fresh projects to ignore these files so they don't accidentally get committed.

---

Made this PR pre-emptively while we test this change internally. We'll likely also have to make a 'clean' command to wipe out files.

### DIFF
--- a/.changeset/modern-points-smash.md
+++ b/.changeset/modern-points-smash.md
@@ -1,0 +1,6 @@
+---
+'create-modular-react-app': patch
+'modular-scripts': patch
+---
+
+Don't delete generated artifacts on lib builds.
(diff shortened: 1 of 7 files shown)


LukeSheard:
What's the case for wanting to run build from multiple processes at the same time? Perhaps it would be better to support synchronous builds within modular itself to avoid parent processes needing to call ```modular build```?

PR #275, opened 2021-04-20
Enable development of npm binaries

LukeSheard:
Currently there are two issues when developing binaries: 

1. Shebangs from files are removed by the current shebang plugin.
2. Packages must expose their `main` field as `cjs`, `es` and types. Which is not likely if the main package is a binary. 

I've abstracted the `patch-package` changes made internally to handle these cases into a PR into modular core so we can remove the patches and start handing developing bins. 

There's a further discussion of whether `bin` should be a type in modular to handle this case directly.

### DIFF
--- a/packages/modular-scripts/src/build.ts
+++ b/packages/modular-scripts/src/build.ts
@@ -23,6 +23,7 @@ import json from '@rollup/plugin-json';
 import resolve from '@rollup/plugin-node-resolve';
 import babel from '@rollup/plugin-babel';
 import commonjs from '@rollup/plugin-commonjs';
+import { preserveShebangs } from 'rollup-plugin-preserve-shebangs';
 
 import * as ts from 'typescript';
 import * as fse from 'fs-extra';
@@ -182,7 +183,7 @@ const typescriptConfig: TSConfig = {};
 async function makeBundle(
   packagePath: string,
   preserveModules: boolean,
-): Promise<boolean> {
+): Promise<[boolean, boolean]> {
   const console = getConsole(packagePath);
 
   const packageJson = packageJsonsByPackagePath[packagePath];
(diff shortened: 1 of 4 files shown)


packages/modular-scripts/src/build.ts
```diff
@@ -197,15 +198,32 @@ async function makeBundle(
       `${packagesRoot}/${packagePath} is marked private, bailing...`,
     );
   }
-  if (!packageJson.main) {
+
+  let compilingBin = false;
+  let main: string;
+  if (!packageJson.main || !packageJson.bin) {
```
threepointone, line 204: This is failing the tests 
```suggestion
  if (!packageJson.main && !packageJson.bin) {
```

packages/modular-scripts/src/build.ts
```diff
@@ -197,15 +198,33 @@ async function makeBundle(
...
+  let compilingBin = false;
+  let main: string | undefined;
+
+  if (packageJson.main) {
+    main = packageJson.main;
+  } else {
+    if (packageJson.bin) {
+      const bins: string[] = Object.values(packageJson.bin) as string[];
+      if (bins.length === 1) {
+        compilingBin = true;
```
threepointone, line 211: dropping this todo for me to handle later 

```suggestion
        // TODO: hoist the logic that determines `compilingBin` outside of makeBundle
        compilingBin = true;
```

packages/modular-scripts/src/build.ts
```diff
@@ -438,49 +441,69 @@ async function makeBundle(
...
+      .replace(path.dirname(main) + '/', '');
+
+    outputFilesPackageJson = {
+      bin: {
+        [binName]: binPath,
+      },
+    };
+  } else {
+    outputFilesPackageJson = {
+      // TODO: what of 'bin' fields?
```
threepointone, line 479:
```suggestion
```

threepointone (approved):
approving with a couple of nits, feel free to land

PR #283, opened 2021-04-21
Expose node.js API for modular-scripts

LukeSheard:
To be able to consume modular-scripts inside the JPM tooling without calling into subprocesses it's useful to be able to call the respective commands natively through node.js. This leaves ```modular``` exposed as a CLI for direct usage but also allows us to leverage modular to higher-level tooling internally.


threepointone:
Did you not want a review on this?

PR #293, opened 2021-04-26
Migrate to commander to enable interactive documentation

LukeSheard:
Screenshot 2021-04-24 at 09 53 55

Migrate the CLI interface to use [commander](https://npm.im/im/commander) to benefit from the CLI interface construction. This exposes better help documentation for each command and adds flag validations.


LukeSheard:
@cangarugula I think this is good to review when you get a chance!

packages/modular-scripts/src/test.ts
```diff
@@ -1,70 +1,71 @@
...
+  // create argv from jest Options
+  const cleanArgv: string[] = [];
 
-  // Watch unless on CI or explicitly running all tests
-  if (!process.env.CI && args.indexOf('--watchAll=false') === -1) {
-    // https://github.com/facebook/create-react-app/issues/5210
-    argv.push('--watchAll');
-  }
+  // pass in path to configuration file
+  const { createJestConfig } = await import('./config/jest');
```
cangarugula, line 61: Why do we need to wait to import this function?

packages/modular-scripts/package.json
```diff
@@ -6,14 +6,17 @@
...
   },
   "scripts": {
     "modular": "ts-node src/cli.ts",
-    "clean": "rimraf build",
+    "clean": "rimraf dist-cjs dist-types",
     "prebuild": "yarn clean",
-    "build": "babel --root-mode upward src --out-dir build --extensions .ts --ignore 'src/**/*.test.ts'"
+    "build": "yarn build:js && yarn build:ts",
+    "build:js": "babel --root-mode upward src --out-dir dist-cjs --extensions .ts --ignore 'src/**/*.test.ts'",
+    "build:ts": "tsc --project tsconfig.build.json"
```
cangarugula, line 19: We can't use modular to build itself?

cangarugula (changes requested):

packages/modular-scripts/src/test.ts
LukeSheard: I wanted to hide away the jest imports given I know it has a deep dependency tree - so this should minimize the import cost of test.

packages/modular-scripts/package.json
LukeSheard: There's an open issue to do this - we don't really support ```bin``` and ```main``` side by side at the moment so it's likely we need some changes to how we build packages.

cangarugula (approved):

PR #333, opened 2021-05-06
Use ```esbuild-loader``` in place of ```babel-loader``` in Create-React-App

LukeSheard:
Replace the transpilation of JavaScript and TypeScript code in ```webpack``` with ```esbuild-loader``` to enable speed benefits from using esbuild directly. I've tied the implementation of ```esbuild``` to the version we use in ```modular-scripts``` so that our release train can match esbuild.


packages/modular-scripts/DO_NOT_IMPORT_THIS_OR_YOU_WILL_BE_FIRED_craco.config.js
```diff
@@ -19,18 +20,39 @@ const absoluteModularGlobalConfigsPath = path.resolve(modularRoot, 'modular');
 
 module.exports = {
   webpack: {
-    configure(webpackConfig) {
-      const { isFound, match } = getLoader(
-        webpackConfig,
-        loaderByName('babel-loader'),
```
padmaia, line 25: Does our previous configuration allow for custom babel configuration or is craco turning that off? Just wondering if this would be considered a breaking change.

padmaia (approved):

packages/modular-scripts/DO_NOT_IMPORT_THIS_OR_YOU_WILL_BE_FIRED_craco.config.js
LukeSheard: Hmm - good point. I suspect then yes this would become a breaking change due to that fact. I would imagine we likely want to allow a very minimal set of babel plugins (or leave ```babel-loader``` present and allow people the option to handle such things themselves? e.g. I guess this would break anything using ```babel-plugin-macros``` right now, but I'm not sure how we'd handle such cases with a ```vite``` migration either. 

@threepointone how did you handle this case with the V2 / ```vite``` migration?

PR #365, opened 2021-05-20
Allow certain jest options to override the set jest config

cangarugula:
- Moved util files into a util folder
- Removed the jest-config and jest-setupEnvironment.js files
- Pulled config logic into its own folder
- Removed coverageThreshold from the craco config

First, it checks if there are any jest options in package.json or in a jest.config.js file. If there are options in both locations, you will see this warning and it will exit. 
Screen Shot 2021-05-20 at 2 28 49 PM

Second, it checks if there are setupFiles and setupFilesAfterEnv options set. If there are, it will instruct the user to move the code to the appropriate modular file. 

Screen Shot 2021-05-20 at 2 03 21 PM

Third, it checks if the options contain keys that are not in the list below and if there is, it will list them as unsupported options. 

```
const supportedOverrides = [
  'collectCoverageFrom',
  'coveragePathIgnorePatterns',
  'coverageThreshold',
  'modulePathIgnorePatterns',
  'testPathIgnorePatterns',
];
```
Screen Shot 2021-05-20 at 2 04 38 PM


packages/modular-scripts/package.json
```diff
@@ -13,7 +13,7 @@
     "modular": "ts-node src/cli.ts",
     "clean": "rimraf build",
     "prebuild": "yarn clean",
-    "build": "babel --root-mode upward src --out-dir build --extensions .ts --ignore 'src/**/*.test.ts'"
+    "build": "babel --root-mode upward src --out-dir build --extensions .ts,.js --ignore 'src/**/*.test.ts'"
```
LukeSheard, line 16: Do we need this? I think we can remove it.

packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -0,0 +1,116 @@
...
+import path from 'path';
+import chalk from 'chalk';
+import type { Config } from '@jest/types';
+import { getCracoJestConfig } from '../index';
+import getModularRoot from '../../utils/getModularRoot';
+import { ModularPackageJson } from '../../utils/isModularType';
+
+const modularRoot = getModularRoot();
+
+const rootPackageJson = fs.readJSONSync(
```
LukeSheard, line 11: Avoid these being in the top level import since they'll run for all commands which will slow startup. Especially blocking IO operations like ```existsSync```.

packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  path.join(modularRoot, 'package.json'),
+) as ModularPackageJson;
+
+const packageJsonJest = rootPackageJson?.jest as
+  | Config.InitialOptions
+  | undefined;
+
+const jestConfigFile = fs.existsSync(path.join(modularRoot, 'jest.config.js'));
```
LukeSheard, line 19: Do we need to check for the existence of a jest config? I think if one exists it should be removed? What's happens with CRA if there's a ```jest.config``` file?

packages/modular-scripts/src/config/index.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
LukeSheard, line 1: What's the reason for this being a js file and now TypeScript?

LukeSheard (changes requested):
Few comments inline - overall looking good!

packages/modular-scripts/src/config/jest/index.ts
cangarugula: Good catch

packages/modular-scripts/src/config/index.js
cangarugula: This started out as the `jest-config.js` file which I reallocated and renamed. I kept it as a js file because craco does not provide types so we'd have to declare @craco as a module and set createJestConfig as an export within modular-scripts. This is why I added the `.js` to the build script. I personally wasn't sure which route to go. Both are not ideal.

cangarugula:
Message if jest.config.js file is found
Screen Shot 2021-05-21 at 12 11 27 PM

packages/modular-scripts/src/config/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+const resolveAsBin = require('resolve-as-bin');
+const path = require('path');
+const { createJestConfig } = require('@craco/craco');
+
+const packagesRoot = 'packages';
+const outputDirectory = 'dist';
+
+const cracoBin = resolveAsBin('craco');
```
LukeSheard, line 10: This is top level again...

packages/modular-scripts/src/config/index.js
```diff
@@ -0,0 +1,28 @@
...
+const cracoConfig = path.join(
```
LukeSheard, line 11: Move this to this folder to live with this..

packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -0,0 +1,121 @@
+import * as fs from 'fs-extra';
+import path from 'path';
+import chalk from 'chalk';
+import type { Config } from '@jest/types';
+import { getCracoJestConfig } from '../index';
+import getModularRoot from '../../utils/getModularRoot';
+import { ModularPackageJson } from '../../utils/isModularType';
+
+// Priority of jest configurations:
```
LukeSheard, line 9: 👋

packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -0,0 +1,121 @@
...
+          "jest": {}
+        }
+        `,
+      ),
+    );
+    process.exit(1);
+  }
+
+  if (packageJsonJest) {
+    const customJest = (
```
LukeSheard, line 66: I think this needs to be tidied up

LukeSheard (approved):

PR #377, opened 2021-05-24
Add babel-preset-react-app to babel-jest config

cangarugula:
Babel-jest needs to transform with react-app presets, similar to react-scripts babelTransformer

### DIFF
--- a/packages/modular-scripts/src/config/jest/index.ts
+++ b/packages/modular-scripts/src/config/jest/index.ts
@@ -38,15 +38,15 @@ export default function createJestConfig(): Config.InitialOptions {
     resetMocks: false,
     transform: {
       '^.+\\.(js|jsx|mjs|cjs)$': [
-        'babel-jest',
+        require.resolve('babel-jest'),
         {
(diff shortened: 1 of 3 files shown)


packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -37,8 +37,13 @@ export default function createJestConfig(): Config.InitialOptions {
     ...defaults,
     resetMocks: false,
     transform: {
-      '^.+\\.jsx?$': 'babel-jest',
-      '^.+\\.tsx?$': 'ts-jest',
+      '^.+\\.(js|jsx|mjs|cjs)$': [
+        'babel-jest',
```
LukeSheard, line 41: Thinking about this it should be ```require.resolve``` so that we use the version that we depend on to make sure that no weird hoisting happens.

cangarugula: Done

LukeSheard:
LGTM! 🚀

PR #379, opened 2021-05-24
Expose extended workspace info via cli

LukeSheard:
Expose extended ```yarn workspaces info``` information embedded with modular information.

### DIFF
--- a/packages/modular-scripts/src/cli.ts
+++ b/packages/modular-scripts/src/cli.ts
@@ -137,6 +137,15 @@ program
     return start(packageName);
   });
 
+program
+  .command('workspace')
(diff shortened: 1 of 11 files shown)


packages/modular-scripts/src/cli.ts
```diff
@@ -137,6 +137,15 @@ program
     return start(packageName);
   });
 
+program
```
padmaia, line 140: Should this new command be added to the README?

LukeSheard: Good shout - added something for the command.

padmaia (approved):

PR #400, opened 2021-05-26
Remove unneccessary silent flag in cmra cli

cangarugula:
I removed the --silent flag from the create-modular-react-app cli.

Create-modular-react-app tests install the latest published version of modular-scripts. The latest published version only allows for 2 flags with the add CLI command (--unstable-name and --unstable-type) and when CMRA uses the --silent flag, it rejects it. 

The --silent flag isn't necessary for CMRA because addPackage already sets the --silent flag when it comes time to yarn install for the new package and modular root.

### DIFF
--- a/packages/create-modular-react-app/src/cli.ts
+++ b/packages/create-modular-react-app/src/cli.ts
@@ -135,7 +135,6 @@ export default function createModularApp(argv: {
       'app',
       '--unstable-name',
       'app',
-      '--silent',
       ...preferOfflineArg,
     ],
     {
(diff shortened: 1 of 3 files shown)


packages/create-modular-react-app/src/cli.ts
```diff
@@ -135,7 +135,6 @@ export default function createModularApp(argv: {
       'app',
       '--unstable-name',
       'app',
-      '--silent',
```
LukeSheard, line 138: Ah I think this is my mistake - this should be the first argument after yarnpkg so that it silences ```yarnpkg``` not modular. I'll cross check this in the morning.

PR #443, opened 2021-06-07
Add more test flags

cangarugula:
Ability to control caching within tests and debugging memory leaks


packages/modular-scripts/src/cli.ts
```diff
@@ -125,6 +125,10 @@ program
     testOptions.watchAll.description,
     !process.env.CI,
   )
+  .option('--bail', testOptions.bail.description)
```
LukeSheard, line 128: Let's default this to true in CI to save CPU time.

packages/modular-scripts/src/cli.ts
```diff
@@ -125,6 +125,10 @@ program
...
+  .option('--clearCache', testOptions.clearCache.description)
```
LukeSheard, line 129: Same with this - default to true in CI.

LukeSheard (changes requested):

packages/modular-scripts/src/cli.ts
cangarugula: We can't default this one to true because it exits the script once it's cleared the cache. It can't be combined with the actual test command or it will never run

LukeSheard: Ah - fair enough then! Maybe it's ```--no-cache``` we want to default to true on CI then.

PR #453, opened 2021-06-10
Modular test ide compatibility

cangarugula:
IntelliJ Jest debugger passes testNamePatterns as a full string (with spaces and regex characters), which we need to parse as a command to maintain the full integrity of the regex and allow for other potential IDE options passed to modular test
 
```
"--testNamePattern=^Test Blotter Settings Dialog renders the grid settings tab by default$" 
```

On the IntelliJ debugger, execa.sync stdout string consoles ANSI escape characters, the command and the "Done in <time> secs" in the stdout string. We need to string everything extraneous before passing it to JSON.parse in `getAllWorkspaces`

Example stdout string that caused JSON.parse errors:
```
'\x1B[2K\x1B[1G\x1B[1myarn workspaces v1.22.10\x1B[22m\n' +
    '\x1B[2K\x1B[1G{\n' +
    ...workspaces
}\n' +
    '\x1B[2K\x1B[1GDone in 0.10s.',
```


packages/modular-scripts/src/utils/getAllWorkspaces.ts
```diff
@@ -12,14 +13,22 @@ type YarnWorkspaces = Record<string, YarnWorkspace>;
 
 function _getAllWorkspaces(): YarnWorkspaces {
   const modularRoot = getModularRoot();
-  return JSON.parse(
+
+  // strip out ANSI color codes and escape characters
+  const strippedStd = stripAnsi(
     execa.sync('yarnpkg', ['workspaces', 'info'], {
```
LukeSheard, line 19: If you run `yarnpkg --silent workspaces info` then it won't add the ansi.

cangarugula: --silent will remove the command and `Done in` but not the ansi characters (in the debugger).  I'll add this in!

PR #486, opened 2021-06-21
Render view types

cangarugula:
Modular start will accept app types and view types to spin up a dev server for.

When pointed to a view type, modular will spin up an app template and import the view as the main app to render. 

This implementation:
- allows for multiple dev servers to be spun up per view
- provides hot reload of the view
- paves the way for the view to built as a standalone app


packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -251,6 +258,18 @@ describe('modular-scripts', () => {
       }
     }
 
+    if (port) {
+      // kill all processes listening to the dev server port
+      exec(
+        `lsof -n -i4TCP:${port} | grep LISTEN | awk '{ print $2 }' | xargs kill`,
```
tstordyallison, line 264: I would guess if we ever add Windows test support this will crop up then...

cangarugula: Yeah - this will need to be updated for Windows test support. We want to redo these tests eventually as well to cleanly start and finish every time.

LukeSheard:
LGTM but should have some docs somewhere?

LukeSheard (approved):

cangarugula:
> LGTM but should have some docs somewhere?

Definitely! In another PR :)

PR #491, opened 2021-06-22
Update vscode configs in template and moved utils

cangarugula:
- Updated the vscode jsons in the CMRA template
- Moved getAllFiles into its own utility
- Pointed tree-view-for-tests test to a more stable folder
- Ignore dist files for formatting


packages/tree-view-for-tests/src/__tests__/tree.test.ts
```diff
@@ -6,33 +6,18 @@ test('it can serialise a folder', () => {
   // so can't include any .test.ts files that actually use this.
   // I picked one of our packages instead.
 
-  expect(tree(path.join(__dirname, '../../../create-modular-react-app')))
+  expect(tree(path.join(__dirname, '../../../modular-views.macro')))
```
cangarugula, line 9: CMRA uses tree in its tests so it causes too many snapshot updates to this test tile

packages/create-modular-react-app/src/__tests__/index.test.ts
```diff
@@ -39,69 +39,65 @@ async function readCensoredPackageJson(
...
+    destination = path.join(tmpDirectory.name, repoName);
+  });
+
+  afterEach(() => {
+    //remove project
+    tmpDirectory.removeCallback();
+    // @ts-ignore
+    destination = undefined;
+  });
+  it('should create a project with defaults', async () => {
```
cangarugula, line 57: I had to pull out these tests into their own individual it block because jest did not like having multiple inline snapshots for the same calls

package.json
```diff
@@ -55,7 +55,7 @@
   },
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint --cache --ext .js,.ts,.tsx --max-warnings 0",
-    "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write"
+    "*.{js,json,ts,tsx,css,md,mdx}": "prettier --write ."
```
LukeSheard, line 58: Why are we formatting the entire project every time?

packages/tree-view-for-tests/src/__tests__/tree.test.ts
```diff
@@ -6,33 +6,12 @@ test('it can serialise a folder', () => {
   // so can't include any .test.ts files that actually use this.
   // I picked one of our packages instead.
 
-  expect(tree(path.join(__dirname, '../../../create-modular-react-app')))
+  expect(tree(path.join(__dirname, '../../../eslint-config-modular-app')))
```
LukeSheard, line 9: I think you'll want to rebase this as I already updated it to point to itself in a meta-testing kind of fashion.

LukeSheard:
Looks fine - can you try rebasing first though as I'm surprised there's no conflicts.

packages/tree-view-for-tests/src/__tests__/tree.test.ts
cangarugula: I think eslint-config-modular-app is a more stable folder to point to. It doesn't have tests with snapshots and tree-view tests does.

package.json
cangarugula: I didn't add in the prettier step in lint-staged but just updated it to run properly. I think formatting as a task will help the repo remain formatted.

packages/tree-view-for-tests/src/__tests__/tree.test.ts
cangarugula: I just checked the last PRs and this change didn't make it in there.

cangarugula:
> Looks fine - can you try rebasing first though as I'm surprised there's no conflicts.
Done

packages/tree-view-for-tests/src/__tests__/tree.test.ts
LukeSheard: Ah - it was this PR which I updated it in: https://github.com/jpmorganchase/modular/pull/492. I think the only issue is for me that these will eventually change so then we'll get "unrelated failures" where as using the tree folder keeps it bound together. I'm happy to merge this though for now.

package.json
LukeSheard: This command (without the `.`) should format any files which are changed though - unless we're missing something. 
I think this version of the command could just go in the pre-commit hook in husky? It's a somewhat bad precedent to set linting the entire codebase - I can imagine a larger modular repo definitely wouldn't want to format 1000s of files in one go!

cangarugula: lint-staged is the precommit husky hook  but yeah I do agree it can get intensive. I re-read the docs and I think I misread them so I will revert

packages/tree-view-for-tests/src/__tests__/tree.test.ts
cangarugula: I'm going to do the proper thing and just set up and tear down a test dir for this...

PR #492, opened 2021-06-22
Fix issues with create-modular-react-app

LukeSheard:
Fixed issue with `create-modular-react-app` need a `<name>` variable.

### DIFF
--- a/packages/create-modular-react-app/src/index.ts
+++ b/packages/create-modular-react-app/src/index.ts
@@ -1,34 +1,157 @@
-#!/usr/bin/env node
+import execa from 'execa';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import chalk from 'chalk';
 
-import createModularApp from './cli';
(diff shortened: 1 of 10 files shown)


packages/create-modular-react-app/src/index.ts
```diff
@@ -1,34 +1,157 @@
...
+    },
+  );
+
+  if (argv.repo !== false) {
+    execSync('git', ['add', '.'], {
+      cwd: newModularRoot,
+    });
+
+    // don't try to commit in CI
+    if (!process.env.CI) {
```
cangarugula, line 149: Checking for CI tests in the source code has a code smell... And I don't know if we should be committing automatically for projects. This will violate the webhook that checks for JIRA tickets in the commits.

LukeSheard: I'm happy to remove the `CI` check - I'm not sure why this would be being used in CI anyway.

PR #498, opened 2021-06-23
Move verifyPackageTree into modular-scripts

LukeSheard:
Move the script which verifies the package installation tree for react-scripts into modular-scripts. This means we can run craco with preflight checks skipped automatically which was previously causing some confusion. 

Secondly this will minimize the gap required when we want to start owning more build configurations.


packages/modular-scripts/src/utils/verifyPackageTree.ts
```diff
@@ -0,0 +1,152 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const ownPackageJson = require('react-scripts/package.json') as PackageJson;
+  const dependencies: Dependency = ownPackageJson.dependencies || {};
+  const expectedVersionsByDep: Record<string, string> = {};
+  // Gather wanted deps
+  depsToCheck.forEach((dep) => {
+    const expectedVersion = dependencies[dep];
+    if (!expectedVersion) {
+      console.log('HMM', dep, expectedVersion);
+      throw new Error('This dependency list is outdated, fix it.');
```
cangarugula, line 43: This seems a bit aggressive for an error...

packages/modular-scripts/src/start.ts
```diff
@@ -22,6 +22,7 @@ export default async function start(appPath: string): Promise<void> {
     env: {
       USE_MODULAR_BABEL: process.env.USE_MODULAR_BABEL,
       MODULAR_ROOT: modularRoot,
+      SKIP_PREFLIGHT_CHECK: 'true',
```
cangarugula, line 25: It seems a bit premature to set this to auto skip now when we haven't taken over build config and still using react-scripts. We can wait until there is a reason to skip preflight checks to set this

LukeSheard: We already call it directly ourselves in `preFlightCheck ` currently - this moves it directly into our startup checks so the gap between react-scripts and the `start` / `build` APIs are smaller for taking control of the build directly.

PR #521, opened 2021-06-30
Modular convert

cangarugula:
- Convert the react app (with src and public folders in the root) into a modular repo with a modular app.
- Conversion is atomic so if it cannot cleanly convert, it will reset the changes.
- Added a root type to modular scripts for modular init to set up a potentially blank project with the modular scaffold. This lends itself to being used by CMRA to initialize itself. 
- check is now validating the properties of the workspaces property in package.json


LukeSheard (changes requested):

packages/modular-scripts/src/convert.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch (err) {
+    logger.error(
+      'Failed to convert your react app to a modular app cleanly. Reverting git changes...',
+    );
+    resetChanges();
+    throw err;
+  }
+}
+
+export default actionPreflightCheck(convert);
```
LukeSheard, line 107: `actionPreflightCheck` runs `check` before running the action - will this not always fail if you run convert on an app which isn't already a modular one?

packages/modular-scripts/types/root/packages/README.md
```diff
@@ -0,0 +1 @@
+This will be the readme inside /packages
```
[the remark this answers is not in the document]
cangarugula: We have to keep it because an empty packages folder will not be picked up by git

packages/modular-scripts/src/convert.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+  return trackedChanged.length === 0;
+}
+
+function resetChanges(): void {
+  execa.sync('git', ['clean', '-fd']);
+}
+
+process.on('SIGINT', () => {
+  logger.error(
```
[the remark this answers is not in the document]
cangarugula: I've interrupted modular start after converting the app and it has not reset git changes.

packages/modular-scripts/src/convert.ts
cangarugula: Yeah you're right. I've always been running convert after init but if someone doesn't they need to know that they should. I'm going to remove actionPreflightCheck.

packages/modular-scripts/types/root/.editorconfig
```diff
@@ -0,0 +1,18 @@
+# https://editorconfig.org
```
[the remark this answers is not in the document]
cangarugula: This is a straight copy over from create-modular-react-app template and it's been fine so far? The only thing I changed was the gitignore into .gitignore. The contents of which are reasonable within the project and items we'd want to ignore anyways.

packages/modular-scripts/src/convert.ts
```diff
@@ -0,0 +1,88 @@
...
+    const packageName = rootPackageJson.name as string;
+
+    // Create a modular app for the current directory
+    await addPackage(packageName, 'app', packageName);
+
+    // Replace the template src and public folders with current react app folders
+    const srcFolders = ['src', 'public'];
+    srcFolders.forEach((dir: string) => {
+      if (fs.existsSync(path.join(cwd, dir))) {
+        rimraf.sync(path.join(cwd, 'packages', packageName, dir));
```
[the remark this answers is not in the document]
cangarugula: I removed addPackage and scaffolded a bare bones modular app instead before moving in the src/public folders

packages/modular-scripts/src/convert.ts
```diff
@@ -0,0 +1,149 @@
...
+  );
+  return trackedChanged.length === 0;
+}
+
+function resetChanges(): void {
+  execa.sync('git', ['clean', '-fd']);
+  throw new Error('Failed to perform action cleanly. Reverting git changes...');
+}
+
+process.on('SIGINT', (err) => {
```
LukeSheard, line 31: I realize this is actually fine due to the async import of the action file.

packages/modular-scripts/src/utils/isModularType.ts
```diff
@@ -38,3 +38,16 @@ export default function isModularType(dir: string, type: PackageType): boolean {
...
 }
+
+export function isValidModularRootPackageJson(dir: string): boolean {
+  const packageJsonPath = path.join(dir, 'package.json');
+  if (fs.existsSync(packageJsonPath)) {
+    const packageJson = fs.readJsonSync(packageJsonPath) as ModularPackageJson;
+    return (
+      packageJson.modular?.type === 'root' &&
+      !!packageJson.private &&
+      !!packageJson?.workspaces?.includes('packages/**')
```
LukeSheard, line 49: This is not technically a requirement, since people could also have `packages/*` or whatever workspace layout they want. We just have convention on `packages/**`. I would argue that these `private` and `workspaces` validations should be in `modular check`.

cangarugula: If they ran modular init, their workspaces will have the `packages/**` convention set as their workspaces. 

This can be placed in modular check but since modular convert doesn't call check until the end, I placed it in a util and called it directly at the start

cangarugula: I added the validation to modular check

packages/modular-scripts/src/check.ts
```diff
@@ -1,14 +1,26 @@
...
 export async function check(): Promise<void> {
   // ensure that workspaces are setup correctly with yarn§
   // init is a special case where we don't already need to be in a modular repository
   // in this case there's no use checking the workspaces yet because we're setting
   // up a new folder
   const workspace = await getWorkspaceInfo();
+  const modularRoot = getModularRoot();
 
+  if (!isValidModularRootPackageJson(modularRoot)) {
+    throw new Error(
```
LukeSheard, line 20: This isn't really a super helpful error - I'll make a PR to make this explicit as to what the real error is.

packages/modular-scripts/src/convert.ts
```diff
@@ -0,0 +1,179 @@
...
+    const newPackagePath = path.join(cwd, 'packages', packageName);
+    fs.mkdirpSync(newPackagePath);
+
+    const newPackageJson = fs.readJsonSync(
+      path.join(packageTypePath, 'packagejson'),
+    ) as ModularPackageJson;
+
+    // Bring key props from root package.json to new app
+    newPackageJson.name = packageName;
+    newPackageJson.modular = {
```
LukeSheard, line 74: @cangarugula I added this - otherwise you won't be able to start the app!

cangarugula: This is already there. We read in the app type packagejson as a JSON (line 68-70) and we update it before writing it again with the updated name and browserlist.

packages/modular-scripts/src/check.ts
cangarugula: If you want to create an issue to have the method itself throw the error, I can raise the PR

packages/modular-scripts/src/convert.ts
LukeSheard: Ah whoops - okay I missed this.

PR #537, opened 2021-07-06
Handle SVG files as URLs instead of components

LukeSheard:
Aligns with the current configuration of `react-scripts`.

See [original config](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L454) for reference. 

Screenshot 2021-07-06 at 13 27 49


packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -37,113 +33,13 @@ function modular(str: string, opts: Record<string, unknown> = {}) {
...
-  };
-}
-
 function cleanup() {
-  rimraf.sync(path.join(packagesPath, 'sample-app'));
   rimraf.sync(path.join(packagesPath, 'sample-view'));
   rimraf.sync(path.join(packagesPath, 'sample-package'));
   rimraf.sync(path.join(packagesPath, 'nested'));
-  rimraf.sync(path.join(modularRoot, 'dist'));
+  rimraf.sync(path.join(modularRoot, 'dist/sample-view'));
```
cangarugula, line 40: why not just delete the dist folder?

packages/modular-scripts/src/__tests__/start-app.tsx
```diff
@@ -0,0 +1,120 @@
+import execa from 'execa';
```
cangarugula, line 1: This should be within a utils folder in `__tests__` so that it's clear it's not a test. Also the extension should be `.ts`

packages/modular-scripts/src/__tests__/index.test.ts
LukeSheard: Because `dist/sample-app` still needs to exist if the tests were to be running at the same time.

cangarugula: We set the tests to run in band sequentially. If this is depending on another test file or another test file is depending on this test file to have a built app, then we should fix that.

packages/modular-scripts/src/__tests__/start-app.tsx
LukeSheard: It's clear it's not a test because it doesn't have the `.test` extension - otherwise we'll end up putting everything in `utils` at this rate!

packages/modular-scripts/src/__tests__/index.test.ts
LukeSheard: In some sense it's not - `/dist` existing is fine because that's our output folder as well.What we really need to do is to fix the tests to run in a directory which is not the the repo!

packages/modular-scripts/src/__tests__/start-app.tsx
cangarugula: And also I just realized there's already a utiils folder within __tests__ so that would not work. Just the extension change then!

packages/modular-scripts/src/__tests__/app.test.tsx
```diff
@@ -0,0 +1,223 @@
+import execa from 'execa';
```
cangarugula, line 1: This should be a `.ts` extension

cangarugula (changes requested):

PR #533, opened 2021-07-01
Build by name instead of location

LukeSheard:
When working with deep nested paths in workspaces then it can be quite tedious to type the full path - this PR changes our approach to build by `name` instead of location from the workspace. This means that it's possible to move apps / libraries around within the workspace without changing build scripts.

PR #539, opened 2021-07-06
Chore: break out buildPackage into separate files

LukeSheard:
This file is mega - for the purpose of mental clarity I've split it out into separate files so that we can move pieces of them in #533


packages/modular-scripts/src/buildPackage/index.ts
```diff
@@ -0,0 +1,154 @@
...
+export async function buildPackage(
+  packagePath: string,
+  preserveModules = false,
+): Promise<void> {
+  const { publicPackageJsons } = getPackageMetadata();
+  const modularRoot = getModularRoot();
+
+  if (process.cwd() !== modularRoot) {
+    throw new Error(
+      'This command can only be run from the root of a modular project',
```
cangarugula, line 40: Won't think be an issue for globally installed modular?

PR #584, opened 2021-07-08
Write workspace info to stdout

cangarugula:
Instead of logging workspace info, it should write it to stdout so that it can be captured


LukeSheard:
Add a changeset and feel free to merge.

LukeSheard (approved):

PR #594, opened 2021-07-12
Modular port <relativePath>

cangarugula:
- Takes a relative path from the modular root to a react app
- Pulls in source content from the targeted app as a modular app in the packages workspace
- Removes react scripts dep and updates react-app-env.d.ts reference with modular-scripts
- Transfers over setupTests file if modular project doesn't already have one in the modular folder
- Resolves dependencies between 2 package.json, giving modular root dependency and dev dependencies precedence
- Removes dependency of a local public package in the workspace to prevent mismatchedWorkspaceDependencies. The ported app will use the symlinked local package's version


cangarugula:
Dependency resolution warnings:
Screen Shot 2021-07-13 at 4 27 37 PM

LukeSheard (approved):

Issue #605, opened 2021-07-14
Disable typechecking in CI


cangarugula:
Currently, typechecking happens every time start, build, or test are executed. For development, this is standard and should be enforced. But for the CI environment, there is no need to typecheck over and over again with every command in the CI pipeline because the code does not change between steps. 

If it's a CI environment, it makes sense to disable typechecking for start, build, and test and make typechecking its own step and done only once over the codebase within the CI environment, preferably before wasting CPU resources to run the other steps if it encounters a failure. This will improve performance of each CI run. 

Typecheck will is still be enabled and enforced during the local development, nothing should change for the everyday development experience. 

a `modular typecheck` command should exist to run typecheck over the project as an explicit action for CI steps.

PR #606, opened 2021-07-14
Disable type check for test, start, build commands in CI

cangarugula:
Typechecks are run every time `start`, `build`, and `test` are run within the CI environment. If it's a CI environment, we should disable typecheck. This will improve performance of each command. 

Because typecheck is enabled and performed during the local development start, build, and test runs, there is no need for the CI to rerun the check with every command in the CI pipeline. Typechecking should be its own step and done only once over the project within the CI environment.

`modular typecheck` runs tsc over the project and should be used explicitly to typecheck the project as part of the SDLC process. 

Addressing #605


packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,13 @@
+import isCI from 'is-ci';
+import execa from 'execa';
+import getModularRoot from './utils/getModularRoot';
+
+export async function typecheck(): Promise<void> {
+  const options = ['--pretty'];
+  if (isCI) {
+    options.push('--noEmit');
+  }
+  await execa('tsc', options, {
```
LukeSheard, line 10: Can we do this programatically like we do for the `build` of packages at the moment?

cangarugula: Done

cangarugula:
For CI:

Screen Shot 2021-07-15 at 2 37 48 PM

For Non-CI
Screen Shot 2021-07-15 at 2 37 24 PM

packages/modular-scripts/src/config/jest/index.ts
```diff
@@ -195,5 +196,16 @@ export function createJestConfig(
       moduleNameMapper: mergedMapper,
     });
   }
+
+  // don't typecheck tests in CI
+  if (isCi) {
+    jestConfig.globals = {
+      'ts-jest': {
+        diagnostics: false,
+        isolatedModules: true,
```
LukeSheard, line 205: I think we always want this to be true? Because it matches the `esbuild-loader` functionality. I thought this was already the case from the `tsconfig.json` we use for `modular-scripts`, but I guess it would be good to enforce this naturally.

packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,65 @@
...
+  const diagnosticHost = {
+    getCurrentDirectory: (): string => ts.sys.getCurrentDirectory(),
+    getNewLine: (): string => ts.sys.newLine,
+    getCanonicalFileName: (file: string): string =>
+      ts.sys.useCaseSensitiveFileNames ? file : toFileNameLowerCase(file),
+  };
+
+  if (diagnostics.length) {
+    if (silent) {
+      throw new Error('\u0078 Typecheck did not pass');
```
LukeSheard, line 52: Can we add a comment as to what this is?

packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,65 @@
...
+  function toFileNameLowerCase(x: string) {
+    return fileNameLowerCaseRegExp.test(x)
+      ? x.replace(fileNameLowerCaseRegExp, x.toLowerCase())
+      : x;
+  }
+
+  const diagnostics = ts.getPreEmitDiagnostics(program) as ts.Diagnostic[];
+
+  const diagnosticHost = {
+    getCurrentDirectory: (): string => ts.sys.getCurrentDirectory(),
```
LukeSheard, line 44: Would the current directory by `modularRoot`?

packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,65 @@
...
+      ts.sys.useCaseSensitiveFileNames ? file : toFileNameLowerCase(file),
+  };
+
+  if (diagnostics.length) {
+    if (silent) {
+      throw new Error('\u0078 Typecheck did not pass');
+    }
+
+    if (isCI) {
+      throw new Error(ts.formatDiagnostics(diagnostics, diagnosticHost));
```
LukeSheard, line 56: Would it be better here do log each of the diagnostics with the modular logger and then throw an error at the end of the failure? This log message will be fairly large.

packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,65 @@
+import isCI from 'is-ci';
+import path from 'path';
+import ts from 'typescript';
+import chalk from 'chalk';
+import getPackageMetadata from './buildPackage/getPackageMetadata';
+import { reportTSDiagnostics } from './buildPackage/reportTSDiagnostics';
```
LukeSheard, line 6: Can we move this to `utils` if we're going to re-use it?

packages/modular-scripts/src/typecheck.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as logger from './utils/logger';
+
+export async function typecheck(silent: boolean): Promise<void> {
+  const { typescriptConfig } = await getPackageMetadata();
```
LukeSheard, line 10: Same with this if we're getting the `typescriptConfig` let's hoist it to be re-usable as we'll probably want to plug it through for `start` / `build` in the future as well?

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -634,8 +634,9 @@ module.exports = function (webpackEnv) {
           // See https://github.com/cra-template/pwa/issues/13#issuecomment-722667270
           maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
         }),
-      // TypeScript type checking
+      // TypeScript type checking for non CI envs
```
LukeSheard, line 637: Can we expand a comment why we're not enabling it for CI environments. Maybe with a link to the GH issue.

LukeSheard (changes requested):
See inline comments - I think a bit more abstraction around typescript configuration generation would be good. Couple of comment expansions would be useful too.

packages/modular-scripts/src/typecheck.ts
cangarugula: How would logging each individually be different than logging them together? Logging each individually would also result in the same long string of type errors to resolve.

packages/modular-scripts/src/typecheck.ts
cangarugula: I'm going to use ts.formatDiagnostic to handle the parsing errors instead

packages/modular-scripts/src/typecheck.ts
cangarugula: Haha yes

packages/modular-scripts/src/config/jest/index.ts
cangarugula: ts-jest has this defaulted to false https://kulshekhar.github.io/ts-jest/docs/getting-started/options/isolatedModules

This is what enables and disables typechecking with ts-jest. I thought we wanted to keep typechecking for local tests?

LukeSheard: Oh I guess this is different to `tsconfig.compilerOptions.isolatedModules` then?

packages/modular-scripts/react-scripts/config/webpack.config.js
cangarugula: Not sure which GH issue you're referring to. This is not part of the memory caching issue for ts-jest if that's what you were referring to but I can link the #605 issue in the comment if that is what you meant?

packages/modular-scripts/src/config/jest/index.ts
cangarugula: Yeah I think this is a ts-jest option that needs to be turned off explicitly

packages/modular-scripts/src/typecheck.ts
LukeSheard: I'm just trying to validate that the output is readable when logging the output in one go vs individually logging the messages. Also it might be good to ensure these are sorted alphabetically!

packages/modular-scripts/react-scripts/config/webpack.config.js
LukeSheard: Yup - let's just add a link to 605.

packages/modular-scripts/src/typecheck.ts
cangarugula: Yeah I'll changing that

packages/modular-scripts/src/typecheck.ts
cangarugula: I can validate that it logs the output in a readable way and alphabetically via path!

packages/modular-scripts/src/cli.ts
```diff
@@ -226,6 +226,22 @@ program
     await port(relativePath);
   });
 
+interface TypecheckOptions {
+  silent: boolean;
+}
+
+program
+  .command('typecheck')
+  .option(
```
LukeSheard, line 235: What's the use case for this? I would avoid adding any unnecessary APIs to this until we've got a better idea of when we'll need them.

LukeSheard (approved):

packages/modular-scripts/src/cli.ts
cangarugula: I added this as a way to silence error messages and provide a green/red light on typechecking. I'm impartial so I will remove for simplicity's sake.

Issue #603, opened 2021-07-14
Default `--preserve-modules` to true 


LukeSheard:
Enable package exports by preserving module structure

PR #608, opened 2021-07-15
#603  Default to `preserve-modules` when building packages

LukeSheard:
- Default `--preserve-modules` to `true` when building packages to enable `export`

field in `package.json` files.

- Fix bug where typings were generated relative to the package directory and not
the `src` directory.

- Add `--private` argument for building packages marked `private: true` in their
`package.json`

### DIFF
--- a/packages/modular-scripts/src/__tests__/index.test.ts
+++ b/packages/modular-scripts/src/__tests__/index.test.ts
@@ -13,6 +13,7 @@ import {
 import getModularRoot from '../utils/getModularRoot';
 import { startApp, DevServer } from './start-app';
 import puppeteer from 'puppeteer';
+import { ModularPackageJson } from '../utils/isModularType';
 
 const rimraf = promisify(_rimraf);
 
@@ -210,7 +211,7 @@ describe('modular-scripts', () => {
           },
           "module": "dist-es/index.js",
           "name": "sample-view",
-          "typings": "dist-types/src/index.d.ts",
+          "typings": "dist-types/index.d.ts",
           "version": "1.0.0",
         }
       `);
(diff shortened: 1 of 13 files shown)


cangarugula (approved):

packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -204,11 +204,11 @@ describe('modular-scripts', () => {
...
           "license": "UNLICENSED",
-          "main": "dist-cjs/sample-view.cjs.js",
+          "main": "dist-cjs/index.js",
           "modular": Object {
             "type": "view",
           },
-          "module": "dist-es/sample-view.es.js",
+          "module": "dist-es/index.js",
           "name": "sample-view",
           "typings": "dist-types/src/index.d.ts",
```
cangarugula, line 213: https://github.com/jpmorganchase/modular/pull/608/files#diff-4870b5ce0bd4c0a47c7dfc47c42422d6e2455d523069b2f0cfbeaf1a70e759b9R33 removed the src folder so you just need update the typings path

cangarugula (changes requested):

cangarugula:
I also noticed that running builds with preserve modules adds the index.js and index.js.map files in dist-cjs but doesn't remove the current app.cjs.js and app.cjs.js.map files in the dir. It's not really an issue for production but this may seem odd when you're testing builds locally with the upgrade. We'll probably note this with the release but the upgrade won't be super tidy.

packages/modular-scripts/src/__tests__/index.test.ts
LukeSheard: Resolved this and added a test case for it!

LukeSheard:
> doesn't remove the current app.cjs.js and app.cjs.js.map files in the dir

I've added `fs.emptyDir` before package builds to match `app` build behaviour so this should be resolved now.

cangarugula (approved):

PR #609, opened 2021-07-15
Remove `bin` discovery for selecting entry files to build.

LukeSheard:
Previously there were two ways to expose an entry point when `modular build <packageName>` was run. In the `package.json` for the directory you could expose `main: "<path to file>` or via a CLI
```{
   "bin": {
      "cli-name": <path to file>
   }
}
```

Modular would prefer `main` over `bin` fields - however the use case for `bin` was redundant given that users could expose via pointing the `bin` field to `dist-{es|cjs}/<path to output>`.

### DIFF
--- a/.changeset/gentle-islands-work.md
+++ b/.changeset/gentle-islands-work.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': major
+---
+
+Remove compiling binaries from "bin" field in package.json.
--- a/packages/modular-scripts/src/__tests__/buildPackage/getPackageEntryPoints.test.ts
+++ b/packages/modular-scripts/src/__tests__/buildPackage/getPackageEntryPoints.test.ts
@@ -1,0 +1,110 @@
+import { getMain } from '../../build/buildPackage/getPackageEntryPoints';
+
+jest.mock('../../utils/getModularRoot', () => {
+  return {
+    __esModule: true,
+    default() {
+      return __dirname;
+    },
+  };
+});
+
+jest.mock('fs-extra', () => {
+  return {
+    __esModule: true,
+    existsSync() {
+      return true;
+    },
+  };
+});
+
+const packagePath = 'modular-scripts';
+
+describe('WHEN given a package.json which doesnt exist', () => {
+  it('SHOULD give an error', () => {
+    expect(() => getMain(packagePath, true, undefined)).toThrow(
+      `no package.json in ${packagePath}, bailing...`,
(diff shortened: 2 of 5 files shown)


packages/modular-scripts/src/__tests__/package.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      │  │  ├─ _rollupPluginBabelHelpers.js #8zp60f
+      │  │  └─ _rollupPluginBabelHelpers.js.map #1a6wdsa
+      │  ├─ dynamic-index.js #k4zvgk
+      │  ├─ dynamic-index.js.map #1shgnup
+      │  ├─ index.js #1jdsfb4
+      │  └─ index.js.map #19flky6
+      ├─ dist-es
+      │  ├─ _virtual
+      │  │  ├─ _rollupPluginBabelHelpers.js #14tvdhd
+      │  │  └─ _rollupPluginBabelHelpers.js.map #4hotzf
```
cangarugula, line 40: What's up with this nested virtual folder?

packages/modular-scripts/src/buildPackage/makeBundle.ts
```diff
@@ -220,11 +225,11 @@ export async function makeBundle(
...
-    throw new Error(
-      `Missing dependencies: ${missingDependenciesWithoutLocalFileNames.join(
-        ', ',
-      )};`,
-    );
+    logger.error(`Missing dependencies found:`);
+    missingDependenciesWithoutLocalFileNames.forEach((missingImport) => {
+      logger.error(`  ${missingImport}`);
+    });
+    throw new Error(`Missing dependencies found.`);
```
cangarugula, line 232: `Missing dependencies found` would be printed twice here.

packages/modular-scripts/src/__tests__/package.test.ts
LukeSheard: It's generated via `@babel/preset-env` for generated helper functions which are used in the output.

packages/modular-scripts/src/buildPackage/makeBundle.ts
LukeSheard: yeah good spot - I'll remove the first comment.

cangarugula:
If this allows modular to build itself, should we make the change to modular-scripts to do this?

Could you add a bit more about why this change is being made? (i.e. What's the value add here?)

PR #659, opened 2021-07-23
Improve `create-modular-react-app` message handling

LukeSheard:
Improve error message handling when creating a new modular repository. Additionally add a `--verbose` flag which proxies to `yarn --verbose` when installing.

### DIFF
--- a/packages/create-modular-react-app/src/cli.ts
+++ b/packages/create-modular-react-app/src/cli.ts
@@ -5,16 +5,23 @@ import { program } from 'commander';
 import * as fs from 'fs-extra';
 import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
 
+// this is a bit gross - but there's no better way of doing this.
+let verbose = false;
+
 // Makes the script crash on unhandled rejections instead of silently
 // ignoring them. In the future, promise rejections that are not handled will
 // terminate the Node.js process with a non-zero exit code.
-// See https://github.com/facebook/create-react-app/blob/f36d61a/packages/react-scripts/bin/react-scripts.js#L11-L16
 process.on('unhandledRejection', (err) => {
-  throw err;
+  if (verbose) {
+    console.error(err);
+  }
+  process.exit(1);
 });
+
 interface Options {
   repo: boolean;
   preferOffline: boolean;
+  verbose: boolean;
 }
 
 program.version(
@@ -28,16 +35,35 @@ program.version(
 program.arguments('<name>');
 program.option('--repo [value]', 'Should a repository be initialized', false);
 program.option('--prefer-offline', 'Yarn --prefer-offline', true);
+program.option('--verbose');
 program.action(async (name: string, options: Options) => {
+  if (options.verbose) {
+    verbose = true;
+  }
+
   await createModularApp({
     name,
     preferOffline: options.preferOffline,
     repo: options.repo,
+    verbose,
   });
 });
 
 try {
-  void program.parseAsync(process.argv);
+  void program
+    .parseAsync(process.argv)
+    .then(() => {
+      process.exit(0);
+    })
+    .catch((err) => {
+      if (verbose) {
+        console.error(err);
+      }
+      process.exit(1);
+    });
 } catch (err) {
-  console.error(err);
+  if (verbose) {
+    console.error(err);
+  }
+  process.exit(1);
 }
--- a/packages/create-modular-react-app/src/cli.ts
+++ b/packages/create-modular-react-app/src/cli.ts
@@ -6,16 +6,13 @@ import * as fs from 'fs-extra';
 import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
 
 // this is a bit gross - but there's no better way of doing this.
-let verbose = false;
+const verbose = process.argv.includes('--verbose');
 
 // Makes the script crash on unhandled rejections instead of silently
 // ignoring them. In the future, promise rejections that are not handled will
 // terminate the Node.js process with a non-zero exit code.
 process.on('unhandledRejection', (err) => {
-  if (verbose) {
-    console.error(err);
-  }
-  process.exit(1);
+  throw err;
 });
 
 interface Options {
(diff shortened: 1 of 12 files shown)


packages/create-modular-react-app/src/index.ts
```diff
@@ -5,21 +5,19 @@ import * as path from 'path';
...
+  console.log(chalk.grey(`$ ${file} ${args.join(' ')}`));
+  try {
+    return execa(file, args, {
+      stdin: process.stdin,
+      stderr: process.stderr,
+      stdout: process.stdout,
+      cwd,
+    });
+  } catch (e) {
+    console.error(chalk.red(`$ FAILED ${file} ${args.join(' ')}`));
```
cangarugula, line 18: Doesn't console.error print in red by default?

packages/create-modular-react-app/src/cli.ts
```diff
@@ -5,16 +5,23 @@ import { program } from 'commander';
...
+// this is a bit gross - but there's no better way of doing this.
+let verbose = false;
+
 // Makes the script crash on unhandled rejections instead of silently
 // ignoring them. In the future, promise rejections that are not handled will
 // terminate the Node.js process with a non-zero exit code.
-// See https://github.com/facebook/create-react-app/blob/f36d61a/packages/react-scripts/bin/react-scripts.js#L11-L16
 process.on('unhandledRejection', (err) => {
-  throw err;
+  if (verbose) {
```
cangarugula, line 15: Given that you catch both async func calls, I don't think this event is likely to happen. But I also think we should just throw the entire err regardless of verbose or not if we do encounter an unhandledRejection error. Let it break so we can fix it.

packages/create-modular-react-app/src/cli.ts
```diff
@@ -5,16 +5,23 @@ import { program } from 'commander';
 import * as fs from 'fs-extra';
 import { JSONSchemaForNPMPackageJsonFiles as PackageJson } from '@schemastore/package';
 
+// this is a bit gross - but there's no better way of doing this.
+let verbose = false;
```
cangarugula, line 9: Could we read it from `program.argv` instead? We allow it as an option so that commander doesn't yell at us and we can pass the option to `createModularApp` in the action but outside of it, we can read `--verbose` from `process.argv`.

packages/modular-scripts/src/cli.ts
```diff
@@ -41,13 +41,15 @@ program
     'Equivalent of --prefer-offline for yarn installations',
     true,
   )
+  .option('--verbose', 'Run yarn commands with --verbose set')
```
cangarugula, line 44: We should set `process.env.MODULAR_LOGGER_DEBUG` to true if they want verbose logs.

packages/modular-scripts/src/cli.ts
```diff
@@ -180,16 +183,22 @@ program
...
   preferOffline: string;
+  verbose: boolean;
 }
 
 program
   .command('init')
   .description('Initialize a new modular root in the current folder')
   .option('-y', 'equivalent to the -y flag in NPM')
   .option('--prefer-offline [value]', 'delegate to offline cache first', true)
+  .option('--verbose', 'Run yarn commands with --verbose set')
```
cangarugula, line 194: We should set `process.env.MODULAR_LOGGER_DEBUG` to true if they want verbose logs.

cangarugula (changes requested):

packages/create-modular-react-app/src/index.ts
LukeSheard: Nope

packages/create-modular-react-app/src/index.ts
```diff
@@ -48,11 +47,13 @@ export default function createModularApp(argv: {
...
 
   const preferOfflineArg = argv.preferOffline ? ['--prefer-offline'] : [];
+  const verboseArgs = argv.verbose ? ['--verbose'] : [];
+
   if (isYarnInstalled() === false) {
     console.error(
       'Please install `yarn` before attempting to run `create-modular-react-app`.',
     );
-    process.exit(1);
+    throw new Error(`Yarn was not installed`);
```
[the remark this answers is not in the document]
LukeSheard: Errors thrown don't the the message logged out anymore - hence the need for the logging and the error here.

packages/modular-scripts/README.md
```diff
@@ -38,7 +38,7 @@ and Tooling together to establish a set of patterns and definitions to enable
 ## Getting Started
 
 ```bash
-  yarn create modular-react-app my-new-modular-project
+  yarn create modular-react-app my-new-modular-project [--verbose]
```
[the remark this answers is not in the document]
LukeSheard: Yup - done!

packages/create-modular-react-app/src/cli.ts
LukeSheard: yeah - I can live with this.

packages/create-modular-react-app/src/cli.ts
LukeSheard: Yeah - seems legit to me. As you said `commander` will validate any args immediately anyway.

packages/modular-scripts/src/cli.ts
LukeSheard: I've updated `MODULAR_LOGGER_DEBUG` to default to `process.argv.includes('--verbose') to match `create-modular-react-app`

packages/create-modular-react-app/src/index.ts
cangarugula: Ah I was thinking of logger.error.

packages/create-modular-react-app/src/cli.ts
```diff
@@ -28,16 +32,31 @@ program.version(
 program.arguments('<name>');
 program.option('--repo [value]', 'Should a repository be initialized', false);
 program.option('--prefer-offline', 'Yarn --prefer-offline', true);
+program.option('--verbose');
```
cangarugula, line 35: Missing a description

cangarugula (approved):

PR #670, opened 2021-07-26
Localize process exits in cli.ts

cangarugula:
Actions should throw error to exit the process. `cli.ts` will control the process interaction and each action can set the message logged via the thrown error or lack thereof.

With the exception of SIGINIT signals for port and convert will call process.exit because there is an additional step taken before exiting.


LukeSheard:
Nice - potentially we should just abstract errors into a `ModularError` though which handles exit codes and messages for printing. Thoughts?

PR #678, opened 2021-07-27
Ensure that errors from verifyPackageTree are surfaced 

LukeSheard:
Adds `esbuild` and `eslint` into `modular-scripts` version management now that we have a `lint` command exposed.


packages/modular-scripts/src/check.ts
```diff
@@ -114,11 +129,153 @@ export async function check(): Promise<void> {
...
+              `P.S. We know this message is long but please read the steps above :-)\n`,
+            ),
+        );
+        failed = true;
+        break;
+      }
+    }
+  }
+
+  console.log();
```
cangarugula, line 259: Delete?

packages/modular-scripts/src/check.ts
```diff
@@ -114,11 +129,153 @@ export async function check(): Promise<void> {
...
+
   try {
-    await verifyPackageTree();
+    require('typescript');
+    throw new Error('lol');
```
cangarugula: lol

packages/modular-scripts/src/cli.ts
```diff
@@ -44,24 +44,26 @@ program
...
+        packageName: string,
+        addOptions: {
+          unstableType?: string;
+          unstableName?: string;
+          preferOffline?: boolean;
+          verbose?: boolean;
+        },
+      ) => {
+        const { default: addPackage } = await import('./addPackage');
+        return addPackage(
```
cangarugula, line 60: addPackage default export is already wrapped with  `actionPreflightCheck`

packages/modular-scripts/src/check.ts
```diff
@@ -114,11 +129,152 @@ export async function check(): Promise<void> {
...
+        break;
+      }
+    }
+  }
+
+  console.log();
+
   try {
-    await verifyPackageTree();
+    require('typescript');
```
cangarugula, line 262: require.resolve?

packages/modular-scripts/src/check.ts
```diff
@@ -114,11 +129,152 @@ export async function check(): Promise<void> {
...
+      if (!fs.existsSync(maybeDepPackageJson)) {
+        return;
+      }
+      const packageJsonString = await fs.readFile(maybeDepPackageJson, 'utf8');
+      const depPackageJson = JSON.parse(packageJsonString) as PackageJson;
+      const expectedVersion = expectedVersionsByDep[dep];
+      if (
+        !semver.satisfies(depPackageJson.version as string, expectedVersion)
+      ) {
+        console.error(
```
cangarugula, line 187: why did you pull this out of `verifyPackageTree`?

packages/modular-scripts/src/cli.ts
```diff
@@ -172,10 +174,12 @@ program
...
     `Start a dev-server for an app. Only available for modular 'app' types.`,
   )
-  .action(async (packageName: string) => {
-    const { default: start } = await import('./start');
-    return start(packageName);
-  });
+  .action(
+    actionPreflightCheck(async (packageName: string) => {
+      const { default: start } = await import('./start');
+      return start(packageName);
```
cangarugula, line 180: start default export is already wrapped with actionPreflightCheck

packages/modular-scripts/src/cli.ts
```diff
@@ -253,24 +257,31 @@ program
...
+  );
 
 void startupCheck()
   .then(() => {
     return program.parseAsync(process.argv);
   })
   .catch((err: Error & ExecaError) => {
     logger.error(err.message);
+    if (logger.DEBUG && err.stack) {
+      logger.error(err.stack);
```
cangarugula, line 284: if `logger.DEBUG` is true, then we should use `logger.debug`

cangarugula:
We should either wrap all of the action exports in `actionPreflightCheck` and remove the wrappings in `cli.ts` or wrap them in cli.ts and remove the default exports that wrap with `actionPreflightCheck`. Let's stay consistent.

packages/modular-scripts/src/check.ts
```diff
@@ -111,14 +114,26 @@ export async function check(): Promise<void> {
...
 
-  /**
-   * Taken from react-scripts - this assets that we share common versions of utils which we depend on.
-   */
   const { verifyPackageTree } = await import('./utils/verifyPackageTree');
+  failed = await verifyPackageTree();
+
   try {
-    await verifyPackageTree();
+    require.resolve('typescript');
```
cangarugula, line 121: This seems like it should belong in verifyPackageTree? If typescript is required, why not put it in `DEPS_TO_CHECK`?

packages/modular-scripts/src/check.ts
```diff
@@ -111,14 +114,26 @@ export async function check(): Promise<void> {
...
   } catch (e) {
     failed = true;
+    console.error(
```
cangarugula, line 124: `logger.error`?

packages/modular-scripts/src/utils/verifyPackageTree.ts
```diff
@@ -33,7 +36,7 @@ export async function verifyPackageTree(): Promise<void> {
   const dependencies: Dependency = ownPackageJson.dependencies || {};
   const expectedVersionsByDep: Record<string, string> = {};
   // Gather wanted deps
-  depsToCheck.forEach((dep) => {
+  DEPS_TO_CHECK.forEach((dep) => {
     const expectedVersion = dependencies[dep];
     if (!expectedVersion) {
       console.log('HMM', dep, expectedVersion);
```
cangarugula, line 42: `logger.log`?

cangarugula (changes requested):

packages/modular-scripts/src/utils/verifyPackageTree.ts
cangarugula: Also, could you make this log a bit clearer on what the issue is? Just "HMM" is a bit passive aggressive lol

packages/modular-scripts/src/check.ts
cangarugula: Or as a dependency?

packages/modular-scripts/src/utils/verifyPackageTree.ts
LukeSheard: This is a *very* unhappy case where we forgot to add a dep to `modular-scripts` `package.json` and therefore can't find the version to match. I've removed it for ease though.

packages/modular-scripts/src/check.ts
LukeSheard: We don't enforce which version of `typescript` you have to use - since it would tie upgrade of `typescript` to `modular-scripts` itself, but rather need to ensure that it's present for command to work properly.

PR #789, opened 2021-09-09
Add unstable esbuild-scripts implementation 

LukeSheard:
Screenshot 2021-09-09 at 17 51 52

Unstable replacement for webpack for application builds using native esbuild APIs and custom reload runtime based on the webpack error overlay.


patches/@svgr+core+5.5.0.patch
```diff
@@ -0,0 +1,107 @@
+diff --git a/node_modules/@svgr/core/index.d.ts b/node_modules/@svgr/core/index.d.ts
+new file mode 100644
+index 0000000..7b54499
+--- /dev/null
++++ b/node_modules/@svgr/core/index.d.ts
+@@ -0,0 +1,101 @@
++export interface TemplateOptions extends SvgrOpts {}
```
cangarugula, line 7: why are you patching these package typings?

cangarugula:
All of the esbuild-scrupt/util files need to be camel cased

packages/modular-scripts/src/build/index.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<void> {
+  const modularRoot = getModularRoot();
+  const targetPath = await getLocation(target);
+
+  await setupEnvForDirectory(targetPath);
+
+  const targetName = toParamCase(target);
+
+  if (isModularType(targetPath, 'app')) {
+    if (process.env.USE_MODULAR_ESBUILD) {
```
cangarugula, line 23: Instead of using process.env, can we add a flatgto the build program? `modular build app --experimental-esbuild`?

patches/@svgr+core+5.5.0.patch
LukeSheard: Since the plugins for esbuild are written in TypeScript to use these modules we need a typings files - I couldn't get it to work with adding them directly into the repo so I thought it easier to patch them into node_modules for typecheck to work.

PR #801, opened 2021-09-13
Move more of react-dev-utils to typescript

LukeSheard:
Move several of the utils from `react-dev-utils` components into the modular CLI so that they are abstracted from the build stage implementation for creating bundles and starting the development server. 

- browserslist configuration auto-updating.
- Required files for application building.
- Hosting instructions.
- File size reporting.
-


packages/modular-scripts/react-scripts/scripts/start.js
```diff
@@ -58,15 +52,7 @@ if (process.env.HOST) {
   console.log();
```
cangarugula, line 52: left over console log?

cangarugula (approved):

PR #835, opened 2021-09-17
Simplify build script + fixes for esbuild compatibility to NPM packages

LukeSheard:

PR #851, opened 2021-09-30
Builds with large webpack stats stall due to writing large JSON

cangarugula:
- Only write the information that we want to bundle-stats.json file
- Expand the linting to beyond the src folder


anuppuranik:
This is a major deal breaker. Gives the impression that something is hung and the modular upgrade hasn't gone well.
Can we have this released sooner than later please. Thanks.

LukeSheard:
Closing in favour of #835

PR #963, opened 2021-10-23
Support svg urls in css files

joshwooding:


joshwooding:
Thanks @LukeSheard for finishing it off 🎉

PR #1025, opened 2021-11-09
Suppress Yarn warnings

cristiano-belloni:
This PR removes the peer dependencies warnings from `create-modular-react-app` and `modular-add`, if the `verbose` flag is not set.

Since [it's impossible to remove warnings from Yarn Classic](https://github.com/yarnpkg/yarn/issues/6672), the suggested solution is just piping the output to grep and remove the offending lines. We can't really assume that the `grep` command (or even a shell to pipe) are available in any environment, though, so this solution uses a stream transform to filter out lines directly within Node.

I found out that `execa` (and `spawn`, which `execa` wraps) can't really pipe stdio to a stream, if the stream [is not backed by a file descriptor](https://nodejs.org/dist/latest-v14.x/docs/api/child_process.html#child_process_options_stdio). For this reason, the caller must be able to provide a custom (optional) configuration to `exec`, and construct the pipeline on its own. Since everything is optional and backwards-compatible, I think it's a fair trade-off.

Screenshots are of `modular add`, CMRA is similar but has more stdout 

Before:

image

After:

image


LukeSheard (approved):

LukeSheard (changes requested):
LGMT - Needs a changeset (run `yarn changeset`) which would probably be a `minor` given this is a feature. 

Also would be good to see a screenshot of the before / after?

cristiano-belloni:
> LGMT - Needs a changeset (run `yarn changeset`) which would probably be a `minor` given this is a feature.

Added.

> Also would be good to see a screenshot of the before / after?

Added.

LukeSheard (approved):

LukeSheard:
Sweet - looks super clean.

PR #1026, opened 2021-11-09
Don't exit too early if there is no modular root.

cristiano-belloni:
- Allow the `--help` and `--version` flags and the `init`, `help` and `convert` commands to work without necessarily being inside a modular directory
- ~Make sure the other commands throw early if there is no modular root~ - not needed, add `actionPreflightCheck` to the `test` command
- Fix a memoizing bug with `convert`, which prevented `getModularRoot` to refresh with the newly-created directory and perform the checks
- Fix `memoize` to support falsy serialized params
- Add tests

### DIFF
--- a/packages/modular-scripts/src/utils/actionPreflightCheck.ts
+++ b/packages/modular-scripts/src/utils/actionPreflightCheck.ts
@@ -1,4 +1,5 @@
 import * as logger from './logger';
+import getModularRoot from './getModularRoot';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type ModularAction = (...args: any[]) => Promise<void>;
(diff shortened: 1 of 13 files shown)


packages/modular-scripts/src/utils/actionPreflightCheck.ts
```diff
@@ -1,11 +1,14 @@
...
+import getModularRoot from './getModularRoot';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type ModularAction = (...args: any[]) => Promise<void>;
 
 function actionPreflightCheck(fn: ModularAction): ModularAction {
   const wrappedFn: ModularAction = async (...args) => {
     if (process.env.SKIP_PREFLIGHT_CHECK !== 'true') {
+      // Bail out if there is no modular root
+      getModularRoot();
```
LukeSheard, line 11: I would put this above the `SKIP_PREFLIGHT_CHECK` so that it always run because otherwise we could get into a situation where people just confused as to what's going go on. The preflight check is really about whether an "actual" modular repo is structurally correct. However we *always* want to make sure we're in a modular repo first.

cristiano-belloni: @LukeSheard - done

LukeSheard (approved):

PR #1033, opened 2021-11-10
Add missing package dependencies and add no-extraneous-dependencies lint rule

### LINKED ISSUE #1031: Setup `import/no-extraneous-dependencies` eslint rule to prevent extraneous dependencies being used in modular.
The default and more common use case for `modular-scripts` is to build with Webpack, switching to `esbuild` by setting the environment variable `USE_MODULAR_ESBUILD` currently throws an exception due to missing dependencies.

The `modular-scripts` package in the `modular` monorepo references `tmp` which is present in the workspace as it's a transitive dependency of the workspace dependency `patch-package`, but it's not loaded in the packaged `modular-scripts`.

Options for solving this include documenting the missing dependencies where they are intentionally excluded from the bundle (listing them as `peerDependencies` would make them easy to find, or include them as part of a readme that documents the environment variable), or explicitly adding missing dependencies to the `dependencies` section of the `package.json`.

Using the `import/no-extraneous-dependencies` lint rule would help ensure future releases contain references to dependencies that could otherwise be missed.

steveukx:
Configure the `import/no-extraneous-dependencies` lint rule to be run against all packages and specify dependencies in sub-packages where they are currently missing as explicit dependencies.

Fixes or adds `eslint-ignore` comments for other linting failures, leaves the project with no `error` level lint failures.

Closes #1031


.eslintrc.js
```diff
@@ -13,8 +13,15 @@ module.exports = {
   extends: ['modular-app'],
   rules: {
     strict: [WARN, 'global'],
+    'import/no-extraneous-dependencies': ['error', { devDependencies: false }],
```
LukeSheard, line 16: I think adding `devDependencies` is fine for the use case outlined in my other comment.

packages/create-modular-react-app/package.json
```diff
@@ -20,6 +20,7 @@
     "build": "babel --root-mode upward src --out-dir build --extensions .ts --ignore 'src/**/*.test.ts'"
   },
   "dependencies": {
+    "@schemastore/package": "0.0.6",
```
LukeSheard, line 23: Hmm - this is technically a false positive. Since this is a "types-only" package then we don't actually want to add it as a dependency. Enabling `devDependencies` and then moving this there should fix that problem.

packages/modular-scripts/package.json
```diff
@@ -90,6 +92,8 @@
     "postcss-preset-env": "6.7.0",
     "postcss-safe-parser": "5.0.2",
     "prompts": "2.4.2",
+    "react": "17.0.2",
```
LukeSheard, line 95: Where do these `react` dependencies come from? We definitely don't want to be shipping a version of `react` / `react-dom` with `modular-scripts`!

packages/modular-scripts/src/esbuild-scripts/plugins/svgr.ts
```diff
@@ -1,7 +1,7 @@
 import * as fs from 'fs-extra';
 import esbuild from 'esbuild';
 
-// @ts-ignore
+// eslint-disable-next-line import/no-extraneous-dependencies
```
LukeSheard, line 4: Does this need to be ignored? I think this should be in `dependencies` list no?

packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -142,12 +142,12 @@ class DevServer {
     });
   }
 
-  shudown = async () => {
+  shutdown = async () => {
```
LukeSheard, line 145: Thanks 😄

LukeSheard (changes requested):

packages/modular-scripts/package.json
LukeSheard: On second thoughts - can we make these `peerDependencies` for `"^16 || ^17"`.

steveukx: It's used throughout the apps under `modular-scripts/types/` - specifying `react` in `dependencies` wouldn't mean it's shipped with `modular`, just that if you `yarn add modular-scripts` without also adding `react` it would be there too... agreed that `peerDependencies` makes more sense though!

packages/create-modular-react-app/package.json
steveukx: Makes sense - I'll move them to `devDependencies` and add inline exceptions

.eslintrc.js
steveukx: I'd prefer to use inline exceptions to the rule rather than allowing `devDependencies` really - enabling `devDependencies` here wouldn't prevent the error the PR is intended to resolve...

packages/create-modular-react-app/package.json
LukeSheard: As you noted before - we can move this to `devDependencies` and add `import type ...` instead.

packages/modular-scripts/package.json
```diff
@@ -121,6 +122,8 @@
     "ws": "7.4.6"
   },
   "peerDependencies": {
+    "react": "17.0.2",
```
LukeSheard, line 125: We shouldn't tie `modular-scripts` to a specific version of `react` etc. Instead we can replace these with `^16 || ^17`.

packages/modular-site/package.json
```diff
@@ -4,10 +4,14 @@
   "license": "Apache-2.0",
   "private": true,
   "dependencies": {
+    "@finos/perspective": "0.6.0",
```
LukeSheard, line 7: I think we can bump these to `0.6.2` to match the current `yarn.lock` versions.

steveukx: I was going for light-touch, so matched it to the version of other `@finos/*` dependencies in `modular-scripts`... do you want them all updating in this package?

packages/modular-scripts/package.json
cristiano-belloni: > On second thoughts - can we make these `peerDependencies` for `"^16 || ^17"` - that's a more accurate representation of them.

@LukeSheard that's tough - peer dependencies mean "something that you need to install in the main package using this plugin / library", which makes sense in a way (if you're using modular, you should definitely install React!) but doesn't in another (this is not a plugin or library that you use at runtime, and anyway React is used only in tests, so it should be a dev dependency).

To be super pedantic, I would put react in dev deps and relax the [rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#options) to not raise an error in certain directories, like `test`, but if you think it's too cumbersome I'm ok with `peerDependencies`.

steveukx: My preference would be to have `react` in `devDependencies` (ie: if you're building/testing `modular` itself, you will get that version of `react` installed), and as a guide to consumers you could list `react` with the broad version in `peerDependencies`, but only if `modular` would always be used in `react` based monorepos.

Due to the way `react` is imported, no errors are thrown by having it specified in `devDependencies`.

packages/create-modular-react-app/package.json
steveukx: I've moved all `@schemastore/` dependencies to `devDependencies` and used `import type...` on the imports themselves.

packages/modular-scripts/package.json
cristiano-belloni: > Due to the way `react` is imported, no errors are thrown by having it specified in `devDependencies`

Why? I'm not incredibly familiar with the rule, but I understood that with `"devDependencies": false` it will error if you import `react` (and we do in `modular-scripts`).

cristiano-belloni: Turns out I didn't see the overrides, which do exactly what I meant two comments ago :) - thanks @steveukx - @LukeSheard if ok for you, I'll close this thread.

yarn.lock
```diff
@@ -9624,8 +9605,8 @@ mkdirp@^0.5.1, mkdirp@^0.5.3, mkdirp@^0.5.5, mkdirp@~0.5.1:
 
 mobile-drag-drop@^2.3.0-rc.2:
   version "2.3.0-rc.2"
-  resolved "https://registry.yarnpkg.com/mobile-drag-drop/-/mobile-drag-drop-2.3.0-rc.2.tgz#00d6e85e04512a620fd5357366e8786bd29aa7aa"
-  integrity sha512-4rHP0PUeWkSp0O3waNHPQZCHeZnLu8bE59MerWOnZJ249BCyICXL1WWp3xqkMKXEDFYuhfk3bS42bKB9IeN9uw==
+  resolved "https://artifacts.jpmchase.net/artifactory/api/npm/npm/mobile-drag-drop/-/mobile-drag-drop-2.3.0-rc.2.tgz#00d6e85e04512a620fd5357366e8786bd29aa7aa"
```
cristiano-belloni, line 9608: @steveukx all these packages have been installed against the internal registry, while modular should use the regular yarn registry (`registry.yarnpkg.com`) - could you 
1) point your registry to the public one
2) clean your yarn cache and node_modules
3) restore the yarn.lock from main
4) re-run `yarn` to update it
5) commit

packages/modular-site/package.json
steveukx: Done!

yarn.lock
steveukx: Thank you for spotting that! This is done

LukeSheard (approved):

PR #1051, opened 2021-11-15
Ignore unsupported browser version in browserlist

cristiano-belloni:
- Still log the error, but ignore an unsupported version of a supported browser in browserlist.
- Added test coverage for `createEsbuildBrowserslistTarget`


LukeSheard (approved):
Sweet! Thanks for the tests as well. 🚀

PR #1060, opened 2021-11-17
Yarn start/build should alias modular start/build and provide the user with an interactive selection if no argument

cristiano-belloni:
- Alias the start script to `modular start` and the build script to `modular build`, so we can use them in the same way (ie: `yarn start my-app`), instead of having them hardcoded to the initial app.
- If `modular start` has no target packages specified, prompt the user to specify a package (see image)

image


packages/modular-scripts/src/start.ts
```diff
@@ -6,14 +6,33 @@ import execAsync from './utils/execAsync';
...
 import * as logger from './utils/logger';
 import createEsbuildBrowserslistTarget from './utils/createEsbuildBrowserslistTarget';
+import prompts from 'prompts';
+
+async function start(packageName: string): Promise<void> {
+  let target = packageName;
+
+  if (!target) {
+    const availablePackages = Object.keys(await getWorkspaceInfo());
+    const chosenTarget = await prompts<string>({
```
cristiano-belloni, line 23: @LukeSheard you mentioned we might want to abstract this in `cli`. Did you mean in `program`, since in `cli` we haven't parsed the arguments / options yet? As I see it, we should still give the possibility to specify a target app to the users (ie: `modular start myApp`) and run the interaction only if no target app is provided (`modular start`), so it should always be post-command.

packages/modular-scripts/src/program.ts
```diff
@@ -56,7 +56,7 @@ program
   );
 
 program
-  .command('build <packages...>')
+  .command('build [packages...]')
```
LukeSheard, line 59: This is still a required argument.

packages/modular-scripts/src/start.ts
LukeSheard: Yeah - I did mean `program` but I think for now what we have is fine - we can abstract this later if we want to re-use it.

packages/modular-scripts/src/program.ts
cristiano-belloni: Good catch! Leftover from my previous build refactor, fixed now.

PR #1063, opened 2021-11-18
Fix modular-scripts builds are failing because of postcss upgrade

cristiano-belloni:
- Bump up postcss-loader and use `implementation` to use postcss 8
- Change the webpack conf to use the new conf schema
- ~Don't lint /config. Not really sure about this, but it seems that modular lint will always die if webpack.config.js is modified and will not let the user commit (because it's invoked by lint-staged), because it tries to run jest on a pattern it doesn't match - what do you think @LukeSheard?~ - reverted, another ticket for this


packages/modular-scripts/src/lint.ts
```diff
@@ -50,7 +50,7 @@ async function lint(
     displayName: 'lint',
     rootDir: modularRoot,
     testMatch: targetedFiles,
-    testPathIgnorePatterns: ['/node_modules/', '/dist/'],
+    testPathIgnorePatterns: ['/node_modules/', '/dist/', '/config/'],
```
LukeSheard, line 53: Let's leave this for now and fix in a different commit by changing the file structure?

cristiano-belloni: If I remove it in the next commit I think it will work - wdym by "changing the file structure"?

cristiano-belloni: ah, ok, you want to move it under src, ok

cristiano-belloni: 👍🏻 Removed, we'll capture this in another ticket

LukeSheard (approved):

PR #1070, opened 2021-11-19
Fix runtime reload

cristiano-belloni:
Fix runtime reload not working in esbuild mode (page not reloading after a file save)

Done:

1. Fix a faulty condition that prevents the page from reloading on a websocket message (runtime thinks there are compile errors when there aren't).
1. Since we're serving the temporary `outDir`, always `write` when watching.
1. Fix the patched definitions of `react-error-overlay` and fix the fact that it excepts early because `default` is undefined.

This doesn't fix the integration with `react-error-overlay` that's still not working and will be tackled in another PR (it errors with `Could not get the stack frames of error: Error: "packages/app/src/App.tsx" is not in the SourceMap` instead of showing the overlay. Source maps are there).

### DIFF
--- a/packages/modular-scripts/src/esbuild-scripts/runtime/index.ts
+++ b/packages/modular-scripts/src/esbuild-scripts/runtime/index.ts
@@ -1,21 +1,25 @@
-import ErrorOverlay from 'react-error-overlay';
+// import {
+//   setEditorHandler,
+//   startReportingRuntimeErrors,
+//   reportBuildError,
+// } from 'react-error-overlay';
 import stripAnsi from 'strip-ansi';
 
 const isFirstCompilation: Record<string, boolean> = {};
 let hasCompileErrors = false;
 
-ErrorOverlay.setEditorHandler(function editorHandler(errorLocation) {
-  // Keep this sync with errorOverlayMiddleware.js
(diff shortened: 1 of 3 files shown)


LukeSheard:
> 2. Understand why bundling the runtime causes a mismatch between the types of `react-error-overlay` and the bundled value. It seems that the type defs and the source have a default export, but our bundled version doesn't. If we import default we get undefined, if we import the individual exports it works but we get a type error. We can easily override the definition, but I'm curious about why it's happening. Might be a weird build on their side? any idea @LukeSheard? (see commented code that would work but doesn't compile)

This is due to the fact the types are patched in via `patch-package` so I could have written them incorrectly. 

https://github.com/jpmorganchase/modular/blob/main/patches/react-error-overlay%2B6.0.9.patch

cristiano-belloni:
> This is due to the fact the types are patched in via `patch-package` so I could have written them incorrectly.
> 
> https://github.com/jpmorganchase/modular/blob/main/patches/react-error-overlay%2B6.0.9.patch

Ah, good to know, thanks. I'll take a look when I come to that part.

LukeSheard (approved):
LGTM 👍

PR #1082, opened 2021-11-22
Fix runtime react error overlay in esbuild mode

cristiano-belloni:
- Use an express middleware to serve the sourcemaps while making the `sources` absolute, for `react-error-overlay` to resolve the sourcemaps correctly. See [this issue](https://github.com/evanw/esbuild/pull/1234) for more info.
- Pass paths to the error formatter to get command-line context for the error.
- Use type imports where possible

---

## Build error in terminal:

image

---

## Build error with REO:

image

---

## Runtime error with REO:

image

---

## REO issue with external directory explanation:

The issue here is that `react-error-overlay` (REO) doesn't find the right source map and dies with an exception instead of showing the stack. It doesn't work with ESBuild and I think it's never worked possibly? (if not, maybe there's another solution?)

Investigating the REO code, it seems that they are using the `source-map` module from Mozilla, which parses the source map generated by ESBuild. When they intercept an error, they try to get the source map content _by the error path_ (which is like `/Users/n761472/dev/test-repo/node_modules/react-dom/cjs/react-dom.development.js`), but since we serve the project from in a `/var/tmp` temp dir, the sources in the sourcemaps are relative to the temp dire, looking something like: `../../../../../../Users/n761472/dev/test-repo/node_modules/react-dom/cjs/react-dom.development.js`. For this reason, `source-map` doesn't match the error (it seems a simple string lookup), throws and the overlay is never shown.

Just changing the `outDir` to our `dist` dir, like in this PR, solves the problem. If we don't have any pressing reason to serve from a temp dir, thjs will work. If I'm not wrong, CRA does it like that - I kinda remember that they serve from `dist`. That would explain why REO just works for them (after all, it's part of CRA). I still have to check if everything works well with `build` and if we're properly emptying `dist` before either buying or serving, but if we choose `dist`, the solution is easy.

If we have reasons to serve from `/var/tmp` instead, we will need to investigate deeper. This is tricky because we're not really under control of what is passed to `source-map` (which is 2 levels deep). We're kinda in control of our source maps, but ESBuild generates them and we can only specify its `sourceRoot`. Maybe setting it to `/`  would work, I don't know. Or maybe there's a way to specify REO init parameters? Or, if all else fails, we can just proxy and modify the sourcemap `source` field when serving?


packages/modular-scripts/src/esbuild-scripts/plugins/wsReload.ts
```diff
@@ -5,6 +5,7 @@ import { formatError } from '../utils/formatError';
 export default function createPlugin(
   name: string,
   server: ws.Server,
+  baseDir: string,
```
LukeSheard, line 8: I would just future-proof this by using `Paths`.

packages/modular-scripts/src/esbuild-scripts/plugins/wsReload.ts
```diff
@@ -20,8 +21,14 @@ export default function createPlugin(
               name,
               building,
               result: {
-                errors: await Promise.all(result.errors.map(formatError)),
-                warnings: await Promise.all(result.warnings.map(formatError)),
+                errors: await Promise.all(
+                  result.errors.map((error) => formatError(error, baseDir)),
```
LukeSheard, line 25: Maybe since formatError is passed around so much we just make it an instance method of the `DevServer` so can just re-use it DRY-ly?

packages/modular-scripts/src/esbuild-scripts/utils/formatError.ts
```diff
@@ -1,11 +1,17 @@
...
 
-export async function formatError(error: Message): Promise<string> {
+export async function formatError(
+  error: Message,
+  baseDir: string = process.cwd(),
+): Promise<string> {
   if (error.location?.file) {
-    const source = await fs.readFile(error.location.file, {
+    const pathToFile = path.join(baseDir, error.location?.file);
+    console.log('Getting error from', pathToFile);
```
LukeSheard, line 13: Is this meant to be here? If so use `logger.debug` from the modular logger to preserve consistent formatting.

LukeSheard:
Apart from the minor comments I made above - I've also been looking at [this issue](https://github.com/evanw/esbuild/pull/1234) about source maps in esbuild and I suspect that we're closer than we think when we ran into this bug around the output generated paths. I think for development middle manning might be the best bet here since there's clearly still some rough edges. 

E.g. in your runtime error screenshot the error path is clearly not correct even though runtime error is pointing to `src/index.ts` as it should be.

Screenshot 2021-11-23 at 21 19 25
 
It might also be nicer if instead of `src/index.ts` it said `packages/foo/src/index.ts` and showed paths relative to the modular root.

cristiano-belloni:
Ah, that makes sense. I can revert to tmp and rewrite the responses via web server. We might even be able to phase it out if / when that pr gets merged.

LukeSheard:
I don't think we need to revert the work so far - I'm more saying that I think the best permanent fix to also take control of the source-map writing, since it's likely we'll want to do this re-writing for production builds as well.

cristiano-belloni:
> I don't think we need to revert the work so far 

By reverting I mean using a tmp directory vs using .modular. Since we will rewrite the source field anyway, there's no difference between the two in the context of using REO. I'm OK with both solutions, both have their pros and cons (I'm not happy of rmrf the .modular subdi every time vs I'm not happy of having to copy paste the random tmp directory if I want to inspect the served bundle). If you think one has more merit than the other, happy to go with it.

packages/modular-scripts/src/esbuild-scripts/plugins/wsReload.ts
cristiano-belloni: Well, it's not passed around, it's `paths` that's passed around, and `formatError` is always imported. We can use a factory (or memoize) to associate an instance of it with a path and pass _that_ around, but I think it'd be messier (we also use it in build, so we'd need two instances). Now that we're using `paths` in our plugins, I think it looks clean.

packages/modular-scripts/src/esbuild-scripts/utils/formatError.ts
cristiano-belloni: Leftover, removed.

cristiano-belloni:
> I think for development middle manning might be the best bet here since there's clearly still some rough edges.

Done. I think that the middleware solution is super clean.

> E.g. in your runtime error screenshot the error path is clearly not correct even though runtime error is pointing to src/index.ts as it should be.

This is solved (see stack trace above)

> It might also be nicer if instead of src/index.ts it said packages/foo/src/index.ts and showed paths relative to the modular root.

Nicer, but that transformation is deep in REO (in the script they dynamically load in the iframe they dynamically create), I don't think we can influence it. The stack trace is relative to the modular root, only the error has directory/filename, I guess it's a UI decision on their end.

PR #1098, opened 2021-11-25
Open IDE from React Error Overlay page links

cristiano-belloni:


packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -28,6 +28,7 @@ import { createIndex } from '../api';
 import { formatError } from '../utils/formatError';
 import createEsbuildConfig from '../config/createEsbuildConfig';
 import { createAbsoluteSourceMapMiddleware } from '../utils/absoluteSourceMapsMiddleware';
+import createLaunchEditorMiddleware from '../../../react-dev-utils/errorOverlayMiddleware.js';
```
LukeSheard, line 31: This is a bit hacky, but probably fine in the interim until we work out a way to bring the build / start code closer together.

LukeSheard (approved):

packages/modular-scripts/src/esbuild-scripts/start/index.ts
cristiano-belloni: Yeah, but we will have to share this middleware in any case. We could move `esbuild-scripts` at the same level of `react-scripts` and a `common-script` directory to store the shared parts, but we would need to invoke `esbuild-scripts` instead of dynamically importing it as we do now. Is there a reason why currently we spawn one but import the other? @LukeSheard?

PR #1103, opened 2021-11-26
Web Worker support for esbuild flow

cristiano-belloni:
This PR introduces an esbuild plugin which:

- Builds workers on the fly, matching them with the `worker.[jt]sx?$` pattern
- Trampolines them to work with CORS (thanks @LukeSheard)
- Emits them as separate files
- Exports a class that needs to be instantiated to start the worker
- Allow only js, ts, jsx and tsx extensions to be loaded from a worker

This plugin works similarly to [worker-loader for Webpack 4](https://v4.webpack.js.org/loaders/worker-loader/) (see below). It will probably superseded [when Esbuild supports `new Worker` as an import](https://github.com/evanw/esbuild/issues/312#issuecomment-800730525), but we will still need trampolining.

## Example

```ts
// index.ts
import HelloWorkerCls from './worker/hello.worker.ts';

const worker = new HelloWorkerCls();
workerRef.current.onmessage = (message) => console.log("Message from worker", message.data);
worker.postMessage(new Date.now());
```

```ts
// ./worker/hello.worker.ts
import { wait, format } from '../utils/date-locale';

globalThis.self.onmessage = async (message: { data: number }) => {
  postMessage(`Hello there. Processing date...`);
  // Simulate work
  await wait(500);
  postMessage(`Date is: ${format(message.data)}`);
};
```

## Caveats

- If workers don't import from other files, they need to `export {}` to not be considered global files by `isolatedModules`.

### DIFF
--- a/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
+++ b/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
@@ -43,7 +43,7 @@ describe('WHEN running esbuild with the workerFactoryPlugin', () => {
       expect(tree(outdir)).toMatchInlineSnapshot(`
         "output
         ├─ alive.worker-T4TLN6IN.js #y0mybi
-        └─ index.js #wd8ozp"
+        └─ index.js #1kx9oa0"
       `);
     });
 
--- a/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
+++ b/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
@@ -6,10 +6,6 @@ import tree from 'tree-view-for-tests';
 import webworkerPlugin from '../../esbuild-scripts/plugins/workerFactoryPlugin';
 import getModularRoot from '../../utils/getModularRoot';
 
-const emptyDir = async (dirName: string) => {
-  await fs.emptyDir(dirName);
-};
-
 describe('WHEN running esbuild with the workerFactoryPlugin', () => {
   describe("WHEN there's a url import", () => {
     let tmpDir: tmp.DirResult;
@@ -34,7 +30,7 @@ describe('WHEN running esbuild with the workerFactoryPlugin', () => {
     });
 
     afterAll(async () => {
-      await emptyDir(tmpDir.name);
+      await fs.emptyDir(tmpDir.name);
       tmpDir.removeCallback();
     });
 
@@ -57,7 +53,7 @@ describe('WHEN running esbuild with the workerFactoryPlugin', () => {
       expect(content).toMatchSnapshot();
     });
 
-    it('SHOULD ouput the correct alive.worker-MLBJYZMX.ts', () => {
+    it('SHOULD ouput the correct alive.worker-[hash].ts file', () => {
       let content = String(
         fs.readFileSync(path.join(outdir, 'alive.worker-T4TLN6IN.js')),
       );
--- a/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
+++ b/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
@@ -10,7 +10,7 @@ const emptyDir = async (dirName: string) => {
   await fs.emptyDir(dirName);
 };
 
-describe('WHEN running esbuild with the svgrPlugin', () => {
+describe('WHEN running esbuild with the workerFactoryPlugin', () => {
   describe("WHEN there's a url import", () => {
     let tmpDir: tmp.DirResult;
     let result: esbuild.BuildResult;
@@ -46,8 +46,8 @@ describe('WHEN running esbuild with the svgrPlugin', () => {
     it('SHOULD have the correct output structure', () => {
       expect(tree(outdir)).toMatchInlineSnapshot(`
         "output
-        ├─ alive.worker-MLBJYZMX.ts #1qkw5er
-        └─ index.js #co48du"
+        ├─ alive.worker-T4TLN6IN.js #y0mybi
+        └─ index.js #18u9kun"
       `);
     });
 
@@ -59,7 +59,7 @@ describe('WHEN running esbuild with the svgrPlugin', () => {
 
     it('SHOULD ouput the correct alive.worker-MLBJYZMX.ts', () => {
       let content = String(
-        fs.readFileSync(path.join(outdir, 'alive.worker-MLBJYZMX.ts')),
+        fs.readFileSync(path.join(outdir, 'alive.worker-T4TLN6IN.js')),
       );
       content = content.replaceAll(getModularRoot(), '');
       expect(content).toMatchSnapshot();
--- a/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
+++ b/packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
@@ -1,0 +1,68 @@
+import * as esbuild from 'esbuild';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import * as tmp from 'tmp';
+import tree from 'tree-view-for-tests';
+import webworkerPlugin from '../../esbuild-scripts/plugins/workerFactoryPlugin';
+import getModularRoot from '../../utils/getModularRoot';
+
+const emptyDir = async (dirName: string) => {
+  await fs.emptyDir(dirName);
+};
+
+describe('WHEN running esbuild with the svgrPlugin', () => {
+  describe("WHEN there's a url import", () => {
+    let tmpDir: tmp.DirResult;
+    let result: esbuild.BuildResult;
+    let outdir: string;
+
+    beforeAll(async () => {
+      tmpDir = tmp.dirSync();
+      outdir = path.join(tmpDir.name, 'output');
+      result = await esbuild.build({
+        entryPoints: [
+          path.join(__dirname, '__fixtures__', 'worker-plugin', 'index.ts'),
+        ],
+        plugins: [webworkerPlugin()],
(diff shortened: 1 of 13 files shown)


LukeSheard:
Few initial thoughts:

1. Maybe the regex should be `/\.worker\.(js|ts|tsx)$/` to ensure it's only files which end with `worker`?
2. We should probably write a plugin which disallows importing anything which is not `.js` `.ts` or `.tsx` files in the web worker otherwise we'll end up with asset files generated from the sub-build. 
2. You'll probably want to trampoline to the web-worker file https://github.com/evanw/esbuild/issues/312#issuecomment-952291570 
3. You should also use `file-loader` to generate the emitted file for the worker so it's included in the build stats. You can probably do this by emitting to in-memory and then using a `onLoad` plugin to read the contents of the build out at the deferred time.

cristiano-belloni:
> 4. You should also use `file-loader` to generate the emitted file for the worker so it's included in the build stats. You can probably do this by emitting to in-memory and then using a `onLoad` plugin to read the contents of the build out at the deferred time.

This is easily done if we want a flow like:

```javascript
import myWorker from './myworker.worker.js'; // imports the hashed filename
// ...
const workerInstance = new Worker(myWorker);
```
Where `myWorker` is simply an emitted hashed worker filename like `myworker.worker-HXDOFRRO.js`
But I guess we want an flow that's identical to Webpack 4 atm, to facilitate ports, correct, @LukeSheard ?

LukeSheard:
> Where myWorker is simply an emitted hashed worker filename like myworker.worker-HXDOFRRO.js 

Sounds perfect to e

> But I guess we want a flow that's identical to Webpack 4 atm, to facilitate ports, correct, @LukeSheard (hence the trampolining)?

I think this is just more about supporting existing patterns and future proofing ourselves against problems we know we'll have. e.g cross-domain CDN asset support is going to be a blocker immediately.

cristiano-belloni:
> 1. Maybe the regex should be `/\.worker\.(js|ts|tsx)$/` to ensure it's only files which end with `worker`?

`/worker-loader:.*\.worker.[jt]sx?$/`

> 2. We should probably write a plugin which disallows importing anything which is not `.js` `.ts` or `.tsx` files in the web worker otherwise we'll end up with asset files generated from the sub-build.

Done

> 3. You'll probably want to trampoline to the web-worker file [WebWorker support? evanw/esbuild#312 (comment)](https://github.com/evanw/esbuild/issues/312#issuecomment-952291570)

Done

> 4. You should also use `file-loader` to generate the emitted file for the worker so it's included in the build stats. You can probably do this by emitting to in-memory and then using a `onLoad` plugin to read the contents of the build out at the deferred time.

Done

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
```diff
@@ -0,0 +1,141 @@
...
+            path.dirname(importAbsolutePath),
+            path.basename(importAbsolutePath).replace(/\.[jt]sx?$/, '.js'),
+          );
+
+          // Path passed to esbuild must be either absolute or fully relative.
+          // If it's absolute, snapshots will fail because the built file will contain an absolute (different) path
+          // We need to make it relative to the current working directory, and add './'
+          // because esbuild doesn't like paths like 'packages/appName/src/...'
+          const relativePath =
+            './' + path.relative(process.cwd(), workerAbsolutePath);
```
LukeSheard, line 35: What is process.cwd() going to be? Does this mean modular root? If so use `getModularRoot` to determine this.

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
```diff
@@ -0,0 +1,141 @@
...
+          throw new Error(`Could not find result for ${args.path}`);
+        }
+      });
+    },
+  };
+
+  return plugin;
+}
+
+function createExtensionAllowlistPlugin(
```
LukeSheard, line 118: Can we split this into a seperate plugin file for consistency with the others?

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
```diff
@@ -0,0 +1,141 @@
...
+  allowedExtensions = ['.js', '.jsx', '.ts', '.tsx'],
+): esbuild.Plugin {
+  return {
+    name: 'worker-factory-plugin',
+    setup(build) {
+      // No lookbehind in Go regexp; need to look at all the files and do the check manually.
+      build.onResolve({ filter: /.*/ }, (args) => {
+        // Extract the extension; if not in the allow list, throw.
+        const extension = path.extname(args.path);
```
LukeSheard, line 127: Should this be an `onLoad` since imports normally follow module convention naming and therefore don't have any extensions.

packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      `);
+    });
+
+    it('SHOULD ouput the correct index.js', () => {
+      let content = String(fs.readFileSync(path.join(outdir, 'index.js')));
+      content = content.replaceAll(getModularRoot(), '');
+      expect(content).toMatchSnapshot();
+    });
+
+    it('SHOULD ouput the correct alive.worker-MLBJYZMX.ts', () => {
```
cangarugula, line 60: update to be generic `alive.worker.ts` file?

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
cristiano-belloni: 👍🏻 done

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
cristiano-belloni: > Should this be an onLoad since imports normally follow module convention naming and therefore don't have any extensions.

The idea was that implicit imports are always allowed, because they fall in our allow list in the default case. But now that this plugin is split and can be reused, I think you're right. The only con is that we lose the `importer` in the error message (unless we pass it in the plugin data, but tbh from a performance perspective I'm not terribly happy to run `onResolve` and `onLoad` for all the files in the worker bundle).

I changed it to use `onLoad` and retouched the error message a bit, lmk what you think.

packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
cristiano-belloni: Updated to `SHOULD ouput the correct alive.worker-[hash].ts file`

packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
```diff
@@ -0,0 +1,68 @@
+import * as esbuild from 'esbuild';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import * as tmp from 'tmp';
+import tree from 'tree-view-for-tests';
+import webworkerPlugin from '../../esbuild-scripts/plugins/workerFactoryPlugin';
+import getModularRoot from '../../utils/getModularRoot';
+
+const emptyDir = async (dirName: string) => {
+  await fs.emptyDir(dirName);
```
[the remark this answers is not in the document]
cristiano-belloni: This was from your commit 😂 -- fixed.

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
cristiano-belloni: @LukeSheard  Solved by recreating the abs path from root + relative path and using it as esbuild entrypoint.

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      build.onResolve({ filter: /.*:__worker-url/ }, (args) => {
+        return {
+          path: args.path.split(':__worker-url')[0],
+          namespace: 'worker-url',
+        };
+      });
+
+      build.onLoad({ filter: /.*/, namespace: 'worker-url' }, (args) => {
+        const result = workerBuildCache.get(args.path);
```
LukeSheard, line 90: We should probably pop this from the Map so that there's no memory leaks.

cristiano-belloni: Done

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
LukeSheard: So rather than throw an error - keep it as `onResolve` and return an error message as part of the resolve hook. This will implicitly make the error appear properly in esbuilds output which is the best of both worlds! 🔥

packages/modular-scripts/src/__tests__/esbuild-scripts/workerPlugin.test.ts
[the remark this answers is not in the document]
LukeSheard: ... I'm okay with being called out ... really, I am .... 🙊

packages/modular-scripts/src/esbuild-scripts/plugins/workerFactoryPlugin.ts
cristiano-belloni: @LukeSheard done!

LukeSheard (approved):
Amazing!! 🚀

PR #1112, opened 2021-11-29
Support yarn v3

joshwooding:
Since modular-scripts and modular-views.macro depend on yarn workspaces and the command has changed in yarn v2+ the code needs to accommodate for that.


joshwooding:
I'm not too sure on how to add tests internally for modular. But I've tested the changes to modular-scripts in https://github.com/jpmorganchase/ui-toolkit/pull/295

LukeSheard:
I've nothing against this - but I think if we're going to add support for multiple package managers we should also considering writing this in a way that will allow us to read `npm workspace` output from v7. 

There's also the question of `yarn` v2 which is just dropped here, perhaps we need to feature detect which strategy to read workspace information based semver major of the `yarn --version` output and default to a `explode` hook which says that the version of yarn for the current worktree is unsupported?

LukeSheard:
> with the same caveat as only with `nodeLinker: node-modules`

If there are setup caveats so this then they need to be added to `modular check` to ensure that users don't get weird error messages if there's incorrect setup.

Otherwise I'm happy to go with this, but adding a check for `v4` when it's eventually released to throw in that condition.

If we add support for NPM then we'll cross that bridge when we get there.

joshwooding:
> > with the same caveat as only with `nodeLinker: node-modules`
> 
> If there are setup caveats so this then they need to be added to `modular check` to ensure that users don't get weird error messages if there's incorrect setup.
> 
> Otherwise I'm happy to go with this, but adding a check for `v4` when it's eventually released to throw in that condition.
> 
> If we add support for NPM then we'll cross that bridge when we get there.

Cool, I'll make the changes.

joshwooding:
@LukeSheard All the changes should be done now? Mind reviewing again?

LukeSheard (approved):

PR #1113, opened 2021-11-29
Esbuild improvements

LukeSheard:
Screenshot 2021-11-29 at 23 02 40

Beginning of a serious tidy up focussing on bringing consistency to the webpack and esbuild experiences. 

1. Fixes for sourcemaps to generate sourcemaps which represent state from the modular root instead of the app root.
2. Improves experience of the error overlay to at least show state within the modular repo instead of relative to the app source to deal with cross package errors
3. Output in source tree is now shown relative to the page root.
4. Moved plugins to be co-located with APIs which require them. 
5. `start` is now entirely in-memory reducing footprint onto disk. 
6. Cleaned up metafile output to be relative to modular root which improves clean-ness of API interaction.
7. 'moves' CSS file generation to be in `/static/css/*` instead of `/static/js`
8. `modular build` now produces hashed asset names for cache busting via esbuild hashing API.

Screenshot 2021-11-29 at 23 05 26


.changeset/orange-bulldogs-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'modular-scripts': patch
+---
+
+WRite files in memory when running an esbuild server to fix issues with file
```
cristiano-belloni, line 5: Typo

cristiano-belloni (approved):
Looks good to me, and very fast (I tried it). For the future, would it be worth to have an (hidden?) switch to dump the metafile somewhere? I was looking at it and it seems very useful to debug the build.

LukeSheard:
@cristiano-belloni We'll need to expose it for the future in some form and generate an `asset-manifest.json` file similar to webpack, but we could also just have a flag for it as well.

cristiano-belloni (approved):

PR #1142, opened 2021-12-03
Upgrade to webpack-dev-server 4

LukeSheard:
Upgrade our internal dev server for the webpack build experience to `webpack-dev-server@4.x` this brings with it a tonne of breaking changes to the internal configurations we needed, but outwardly there should be no breaking changes from an API point of view since it's still running `webpack@4` internally.


packages/modular-scripts/react-dev-utils/WebpackDevServerUtils.js
```diff
@@ -177,7 +179,7 @@ function createCompiler({
             'Files successfully emitted, waiting for typecheck results...',
           ),
         );
-      }, 100);
+      }, 0);
```
cristiano-belloni, line 182: Why do we have a timeout?

packages/modular-scripts/react-dev-utils/logger.js
```diff
@@ -0,0 +1,53 @@
...
+const DEBUG =
+  process.env.MODULAR_LOGGER_DEBUG || process.argv.includes('--verbose');
+const SILENT = process.env.MODULAR_LOGGER_MUTE;
+const isInteractive = process.stdout.isTTY;
+
+function printStdErr(x) {
+  if (SILENT) {
+    return;
+  }
+  x.split('\n').forEach((l) => {
```
cristiano-belloni, line 17: Wouldn't `.split(/\r?\n/)` be more universal?

packages/modular-scripts/react-dev-utils/WebpackDevServerUtils.js
LukeSheard: In most cases the diagnostics should come back pretty quickly - so this timeout is in place to ensure we give TypeScript sufficient time to process the files.

packages/modular-scripts/react-dev-utils/logger.js
LukeSheard: Yeah - when I wrote this initially in the main `modular-scripts` package I didn't consider it - I'll update both in place.

cristiano-belloni (approved):
Lgtm

PR #1150, opened 2021-12-06
Replace babel with esbuild for package compilation

LukeSheard:
This replaces the `rollup-plugin-babel` with `rollup-plugin-esbuild` when building libraries.

We generate the same target from the `browserlists` configuration for the repository so that output from libraries and applications support the same browser range as each other. This also reduces the overhead of installation by removing a lot of the babel dependency tree that we had previously.


cristiano-belloni (approved):
👍🏻

cristiano-belloni (approved):
LGTM

cristiano-belloni (changes requested):
I had approved it, but it seems tests are failing

cristiano-belloni (approved):

PR #1152, opened 2021-12-06
Remove service worker impl generation from workbox plugin.

LukeSheard:
Remove support for the workbox plugin as we don't generate service workers by default and they are not important when it comes to our dynamic view loading architecture.


cristiano-belloni (approved):
Agree!

PR #1154, opened 2021-12-06
Bump `tsconfig.json` target to `es2016`

LukeSheard:


cristiano-belloni (approved):
It seems [we don't lose anything](https://kangax.github.io/compat-table/es2016plus/), lgtm

PR #1155, opened 2021-12-06
Simplify package.json scripts

LukeSheard:


.github/workflows/release.yml
```diff
@@ -26,12 +26,15 @@ jobs:
...
         run: yarn --frozen-lockfile
 
+      - name: Build
+        run: yarn build
+
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@master
         with:
           # This expects you to have a script called release which does a build for your packages and calls changeset publish
```
cristiano-belloni, line 36: This comment is now outdated

package.json
```diff
@@ -15,7 +15,6 @@
     "modular": "ts-node packages/modular-scripts/src/cli.ts",
     "test": "yarn modular test --watchAll false --runInBand --forceExit",
     "build": "yarn workspace create-modular-react-app build && yarn workspace modular-scripts build && yarn workspace modular-views.macro build",
-    "release": "yarn build && changeset publish",
```
cristiano-belloni, line 18: Won't this allow people to publish in error without having built?

.github/workflows/release.yml
LukeSheard: Removed 👍

package.json
LukeSheard: I think only the publishing account we created has access to publish - we shouldn't be publishing locally anyway.

cristiano-belloni (approved):

PR #1163, opened 2021-12-08
modular rename command

cristiano-belloni:
**Done:**

1. ~Move old package directory to new package directory~ - We decided with @cangarugula that we don't want to do this as it's ambiguous and directory and package name are not necessarily coupled - see also [docs/commands/rename.md](https://github.com/jpmorganchase/modular/pull/1163/files#diff-d22920075e8cd6c3048519051cb037109437d6342022cec3c867dd2867ad2deb)
1. Change package name in `package.json`
1. ~Change dependency name in depending packages' `package.json`~ - not needed.
1. Re-write imports in depending packages' source files
1. Add logs
1. Write tests
1. Add docs
1. Made action atomic


packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+        .map(async ([_, packageData]) => ({
+          packageData,
+          json: (await fs.readJson(
+            path.join(getModularRoot(), packageData.location, './package.json'),
+          )) as PackageJson,
+        })),
+    )
+  ).filter((pkgJsonInfo) =>
+    dependencyTypes.some(
+      (depField) => pkgJsonInfo.json[depField]?.[oldPackageName],
```
cangarugula, line 87: What's the use case here? If a package is using another package in the same monorepo, they shouldn't need to specify it as a dependency in their package.json. The one case I can think of is if they want to specify an older version of the dependency, in which case, you wouldn't want to change it to the new package name. Maybe I'm missing a use case here.

.gitignore
```diff
@@ -9,6 +9,9 @@ packages/sample-esbuild-app
 packages/sample-view
 packages/sample-package
 packages/nested
+packages/sample-library-package
+packages/sample-depending-package
+packages/sample-library-renamed-package
```
cangarugula, line 14: let's add a scoped and nested example - which will be the majority of apps for our teams

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+    throw new Error(`Package ${oldPackageName} not found.`);
+  }
+  logger.debug(`Checking for collision with ${newPackageName} in workspace.`);
+  if (workspace.find(([packageName]) => packageName === newPackageName)) {
+    throw new Error(`Package ${newPackageName} already exists.`);
+  }
+
+  // Rename the directory. Do it first because this fails if there's a collision and we don't want to clean up later
+  const newPackageLocation = path.join(
+    getModularRoot(),
```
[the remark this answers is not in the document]
cristiano-belloni: @cangarugula It's memoized to allow calling it without performance penalty. I can hoist it, but everywhere else we use the "just call it" pattern, which imho is a bit more readable.

packages/modular-scripts/src/rename.ts
cristiano-belloni: > If a package is using another package in the same monorepo, they shouldn't need to specify it as a dependency in their package.json

Don't they need to specify the dependency in their package.json to get the other package symlinked? I'm fairly new to yarn workspaces but that's what the documentation example says: https://classic.yarnpkg.com/lang/en/docs/workspaces/#toc-how-to-use-it (note that's not necessary to build `package-a`, if you don't need it outside the monorepo you just point `main` or `module` or `exports` to `src`)

There's always the case that `package-b` just imports from `package-a`'s source using a relative path (eg `import libA from "../package-a/src/index.js"`), but then what's the advantage of having workspaces (decoupling packages etc)?

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+    path.join(oldPackage.location, '..'),
```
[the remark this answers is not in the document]
cristiano-belloni: Good point, that needs a bit more work.

packages/modular-scripts/src/rename.ts
cangarugula: They don't need to specify the dependency to use the package immediately. Because we set `"workspaces": ["packages/**]` and we call "yarn" after we create a new package with `/packages`, yarn creates a symlink to the local package in your node_modules, which can be imported/required using its package name. 

But they would need to add it explicitly is if they wanted to externalize the dependency from the build, which case would be to published version that can be installed from the registry. But I can also see the value of doing it anyways just in case they do have the dependency there.

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+            logger.debug(
+              `Rewriting \`${importDeclaration.getText()}\` in ${sourceFile.getFilePath()}`,
+            );
+            importDeclaration.setModuleSpecifier(newPackageName);
+          }
+        });
+      });
+      await project.save();
+    }),
+  );
```
cangarugula, line 147: If you change the package name in the package.json, you'll need to do a `execa.sync('yarnpkg')` to update the the name of the symlink in node_modules

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+async function rename(
+  oldPackageName: string,
+  newPackageName: string,
+): Promise<void> {
+  const workspace = Object.entries(await getWorkspaceInfo());
+
+  logger.debug(`Checking for existence of ${oldPackageName} in workspace.`);
+  const oldPackage = workspace.find(
+    ([packageName]) => packageName === oldPackageName,
+  )?.[1];
```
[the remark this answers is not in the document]
cristiano-belloni: Done

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (!oldPackage) {
+    throw new Error(`Package ${oldPackageName} not found.`);
+  }
+  logger.debug(`Checking for collision with ${newPackageName} in workspace.`);
+  if (workspace.find(([packageName]) => packageName === newPackageName)) {
```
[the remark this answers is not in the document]
cristiano-belloni: Done

packages/modular-scripts/src/rename.ts
cristiano-belloni: Done

packages/modular-scripts/src/rename.ts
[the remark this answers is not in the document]
cristiano-belloni: Done, not renaming directories as discussed.

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,150 @@
...
+
+type DependencyObject = {
+  [key in DependencyType]: Record<string, string>;
+};
+
+interface PackageJson extends Partial<DependencyObject> {
+  name: string;
+}
+
+async function rename(
```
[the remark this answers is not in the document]
cristiano-belloni: Done

packages/modular-scripts/src/rename.ts
cristiano-belloni: Done, removed renaming of dependencies in `package.json`s (I was used to rushjs, where you have to specify internal dependencies. Good to know!)

.gitignore
cristiano-belloni: Now that we don't actually rename directories, do we still need this? Wouldn't like to pollute the repo too much.

.changeset/nasty-mayflies-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"modular-scripts": minor
+---
+
+Add "modular rename" command
```
cangarugula, line 5: Let's add a tiny summary of what rename will do.

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,107 @@
...
+
+type DependencyType = typeof dependencyTypes[number];
+
+type DependencyObject = {
+  [key in DependencyType]: Record<string, string>;
+};
+
+interface PackageJson extends Partial<DependencyObject> {
+  name: string;
+}
```
cangarugula, line 32: You should also be able to get those dependencty types from `import { CoreProperties } from '@schemastore/package'`

.changeset/nasty-mayflies-sin.md
cristiano-belloni: Done, but still really tiny (a bigger description will be added in the next release notes)

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,107 @@
...
+  const workspace = await getWorkspaceInfo();
+
+  if (!cleanGit(process.cwd())) {
+    throw new Error(
+      'You have unsaved changes. Please save or stash them before we attempt to rename a package.',
+    );
+  }
+
+  try {
+    logger.debug(`Checking for existence of ${oldPackageName} in workspace.`);
```
cangarugula, line 47: We should check for the existence and collisions before we ask people to stash their changes.

packages/modular-scripts/src/rename.ts
```diff
@@ -0,0 +1,107 @@
...
+              importDeclaration.setModuleSpecifier(newPackageName);
+            }
+          });
+        });
+        await project.save();
+      }),
+    );
+
+    logger.log(`Refreshing packages`);
+    execa.sync('yarnpkg');
```
cangarugula, line 100: Let's silence the install and specific the cwd as modular root in case they're not in the root. `execa.sync('yarnpkg', ['--silent'], { cwd: modularRoot })`

packages/modular-scripts/src/program.ts
```diff
@@ -253,4 +253,13 @@ program
     await typecheck();
   });
 
+program
+  .command('rename <old-package-name> <new-package-name>')
```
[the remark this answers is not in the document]

packages/modular-scripts/src/rename.ts
cristiano-belloni: Good catch, this was handy before, where I iterated on the depTypes const array and constructed the types from it, but it's not needed now. Done.

packages/modular-scripts/src/rename.ts
cristiano-belloni: Done

packages/modular-scripts/src/__tests__/__fixtures__/packages/sample-depending-package/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+import * as ReactDOM from 'react-dom';
+// @ts-ignore
+import App from './App';
+
+ReactDOM.render(
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+  document.getElementById('root'),
+);
```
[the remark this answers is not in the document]
cristiano-belloni: We still need the `index.tsx`, what we're changing is `App.tsx`. It's true that theoretically we can do everything in one file, without needing a "depending" fixture at all, but I think it's worth showing that, of a project with two files, one file is left untouched and the other one with the dependency is parsed / processed.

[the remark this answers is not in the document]
cristiano-belloni: (btw - removed `sample-library-renamed-package` from the ignore file)

cangarugula (approved):

PR #1193, opened 2021-12-14
Bump Typescript, @typescript-eslint, eslint-plugin-jest, eslint-plugin-testing-library and update tests and definitions

cristiano-belloni:
This PR supersedes https://github.com/jpmorganchase/modular/pull/1186 and https://github.com/jpmorganchase/modular/pull/1166

This PR originated from a bug noticed by @cangarugula, where our Typescript svg definitions were not working. I discovered that the problem happens only with `typescript@>4.5.0 <=4.5.2`. Additionally, I discovered that the tilde definition of our typescript dependency conflicts with `@typescript-eslint` (`yarn lint` printing warnings), now that `4.5`  is out.

There are two possible ways of fixing this: the easy one is constricting our `typescript` dependency to be strictly less than `4.5.0` and dealing with excluding the faulty versions later (https://github.com/jpmorganchase/modular/pull/1166/files). The longer one is to upgrade typescript, force it to be greater or equal to than `4.5.3` to exclude the faulty versions and update all the incompatible dependencies (this PR).

This PR updates:
- `@typescript-eslint` to the latest version + fixes many `any` definitions that are not allowed any more
- `eslint-plugin-jest, eslint-plugin-testing-library` to the latest version + fixes test expectations to align with it (eg `toEqual` becomes `toBe` etc)

This is breaking, since we're bumping `modular-scripts` to use a stricter linter that can break builds and tests.

### DIFF
--- a/.changeset/twenty-cameras-develop.md
+++ b/.changeset/twenty-cameras-develop.md
@@ -1,0 +1,9 @@
+---
+"create-modular-react-app": major
+"eslint-config-modular-app": major
+"modular-scripts": major
+"modular-site": patch
+"modular-views.macro": patch
+---
+
+Bump Typescript, @typescript-eslint, eslint-plugin-jest, eslint-plugin-testing-library and update tests and definitions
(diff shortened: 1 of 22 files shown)


LukeSheard:
@cristiano-belloni what's the resolution plan for this? Was this resolved with `4.6` release of TypeScript or should we close this?

PR #1285, opened 2022-01-13
Generate dependency manifest for apps

cristiano-belloni:
- Extract dependency ranges for a package, using `ts-morph`. This implementation is independent from the build process, and could technically easily offer dependency extraction as a separate command, if needed.
- Write extracted dependencies to `dependencies` and `bundledDependencies` and copy `package.json` to `dist`

This implementation assumes no nested workspaces (Yarn classic), so the dependencies can only be in the package's package.json or in the root one.


packages/modular-scripts/src/build/dependencyManifest.ts
```diff
@@ -0,0 +1,66 @@
...
+    (
+      fs.readJSONSync(
+        path.join(targetLocation, 'package.json'),
+      ) as CoreProperties
+    ).dependencies || {};
+
+  const manifest = getDependenciesFromSource(
+    targetLocation,
+  ).reduce<DependencyManifest>((manifest, depName) => {
+    const depVersion =
```
LukeSheard, line 56: This would be cleaner by creating the dependencies by overwriting from the `targetLocation` into the root dependencies. This would create a "workspace" set of dependencies.

packages/modular-scripts/src/build/dependencyManifest.ts
```diff
@@ -0,0 +1,66 @@
...
+      targetPackageJsonDependencies[depName] ??
+      rootPackageJsonDependencies[depName];
+    if (depVersion) {
```
LukeSheard, line 59: This should throw an error if you're importing something which is not listed as a dependency. ESlint should also pick it up, but people do weird things...

packages/modular-scripts/src/build/dependencyManifest.ts
cristiano-belloni: Done

packages/modular-scripts/src/build/dependencyManifest.ts
cristiano-belloni: Done

packages/modular-scripts/src/build/dependencyManifest.ts
cristiano-belloni: @LukeSheard not sure I understand. Does this mean initialise the reducer with the target dependencies, then add the root dependencies as we iterate through them?

packages/modular-scripts/src/__tests__/app.esbuild.test.ts
```diff
@@ -84,6 +84,7 @@ describe('when working with an app', () => {
       ├─ logo192.png #1nez7vk
       ├─ logo512.png #1hwqvcc
       ├─ manifest.json #19gah8o
+      ├─ package.json
```
LukeSheard, line 87: It would probably be good to include the hash in the test, or at least add a test case for the package.json content.

packages/modular-scripts/src/build/index.ts
```diff
@@ -129,6 +131,20 @@ async function buildApp(target: string) {
     }
   }
 
+  // Add dependencies from source and bundled dependencies to target package.json
+  const packageDependencies = await getPackageDependencies(target);
+  const targetPackageJson = (await fs.readJSON(
```
LukeSheard, line 136: I would suggest for an app that we generate a "minimal" `package.json` content instead of pulling everything directly across. Similar to how we do for `package` types.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -0,0 +1,83 @@
...
+import getWorkspaceInfo from './getWorkspaceInfo';
+
+type DependencyManifest = NonNullable<CoreProperties['dependencies']>;
+
+const npmPackageMatcher =
+  /^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*/;
+
+/* Get dependencies from import / require declarations, since they could be hoisted to the root workspace. Exclude test files. */
+function getDependenciesFromSource(workspaceLocation: string) {
+  const project = new Project();
```
LukeSheard, line 16: Does this need a working directory?

cristiano-belloni: No, you just add the files you want to act on with `addSourceFilesAtPaths`. You can also point it to a `tsconfig`, but I found it gets confused with workspaces.

packages/modular-scripts/src/build/index.ts
```diff
@@ -138,10 +137,20 @@ async function buildApp(target: string) {
...
   targetPackageJson.bundledDependencies = Object.keys(packageDependencies);
-  // Copy package.json over
+  // Copy selected fields of package.json over
   await fs.writeJSON(
     path.join(paths.appBuild, 'package.json'),
-    targetPackageJson,
+    {
+      name: targetPackageJson.name,
+      version: targetPackageJson.version,
+      main: targetPackageJson.main,
```
LukeSheard, line 146: Where should this point once the app has been built? Same for `module` and `typings`?

packages/modular-scripts/src/build/index.ts
```diff
@@ -138,10 +137,20 @@ async function buildApp(target: string) {
...
+      license: targetPackageJson.license,
+      modular: targetPackageJson.modular,
+      files: targetPackageJson.files,
```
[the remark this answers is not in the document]
cristiano-belloni: CRA omits this, which makes sense, since apps don't need importing / requiring

packages/modular-scripts/src/build/index.ts
cristiano-belloni: CRA omits this, which makes sense, since apps don't need importing / requiring

packages/modular-scripts/src/build/index.ts
cristiano-belloni: Done, limiting the fields in the public package.json

packages/modular-scripts/src/__tests__/app.esbuild.test.ts
cristiano-belloni: Done

packages/modular-scripts/src/build/index.ts
```diff
@@ -129,6 +130,27 @@ async function buildApp(target: string) {
...
+  targetPackageJson.bundledDependencies = Object.keys(packageDependencies);
+  // Copy selected fields of package.json over
+  await fs.writeJSON(
+    path.join(paths.appBuild, 'package.json'),
+    {
+      name: targetPackageJson.name,
+      version: targetPackageJson.version,
+      license: targetPackageJson.license,
+      modular: targetPackageJson.modular,
+      module: targetPackageJson.module,
```
LukeSheard, line 148: This will be undefined, but probably best to remove?

cristiano-belloni: Done

LukeSheard (approved):

PR #1290, opened 2022-01-15
Respect original tsconfig exclude when type checking

joshwooding:


packages/modular-scripts/src/utils/getPackageMetadata.ts
```diff
@@ -68,11 +68,8 @@ async function getPackageMetadata() {
     throw new Error('Failed to load Typescript configuration');
   }
   Object.assign(typescriptConfig, configObject, {
-    // TODO: should probably include the original exclude in this
     exclude: distinct([
       // all TS test files, regardless whether co-located or in test/ etc
-      '**/*.stories.ts',
```
joshwooding, line 74: Not sure why wouldn't want to typecheck stories?

LukeSheard:
I think there's potential for improvement in this space when it comes to the `typecheck` inclusion / exclusion. 

Can you provide a little more colour here as to why this is required? What's different about your setup that you need to override the defaults. `modular` commands are supposed to provide a opinionated consistency without much (if any) configuration so this is an "anti pattern" for the direction that we'd probably want to take things.

joshwooding:
> I think there's potential for improvement in this space when it comes to the `typecheck` inclusion / exclusion.
> 
> Can you provide a little more colour here as to why this is required? What's different about your setup that you need to override the defaults. `modular` commands are supposed to provide a opinionated consistency without much (if any) configuration so this is an "anti pattern" for the direction that we'd probably want to take things.

Well, for one since we use typescript for our storybook stories we would want to type check so we can ensure no broken code slips through.

 Plus we want to exclude type checking our cypress tests which use a different file name pattern than our jest tests so we can easily differentiate them.

 This behaviour of silently overriding user config is not transparent to users and can cause a lot of confusion and provides a false sense of security since all of their files aren’t type checked e.g. It took me a while to see that stories were being excluded inside modular and I had to turn on debug to check for certain. If we can’t use modular to type check them we will just run tsc ourselves like we currently do with our tests since there is no way I can see to do this with modular.

packages/modular-scripts/src/typecheck.ts
```diff
@@ -13,16 +13,7 @@ async function typecheck(): Promise<void> {
...
-      'node_modules',
-      'bower_components',
-      'jspm_packages',
-      'tmp',
-      '**/dist-types',
-      '**/dist-cjs',
-      '**/dist-es',
-      'dist',
-    ],
+    exclude: [...(rest.exclude as string[]), 'dist'],
```
joshwooding, line 16: A lot of these are provided by getPackageMetadata.

PR #1382, opened 2022-02-08
Build ESM views with external CDN support

cristiano-belloni:
This is related to https://github.com/jpmorganchase/modular/discussions/1311

- Build views as ESM modules with external CDN
- Build a synthetic `index.html` and `_trampoline.js` to distribute views as static site (`modular serve view` works)
- Support many CDNs with environment templates - for example, `EXTERNAL_CDN_TEMPLATE="https://esm.sh/[name]@[version]"` for esm.sh, `EXTERNAL_CDN_TEMPLATE="https://cdn.skypack.dev/[name]@[version]"` for Skypack. Default is Skypack.
- Support blocklist for external dependencies (can be transformed into allowlist for gradual introduction)
- Set `module` field in output package.json with the built (hashed) module name.
- Set correct `bundledDependencies` array.
- `view serve` works for views
- Use [css module scripts](https://web.dev/css-module-scripts/) for `esnext` targets and fallback to `link` css injection for global css imports from the CDN.
- Fix openBrowser.ts to actually work on CI (wrong default value to satisfy type was forcing it to open OSX Chrome)
- Test all functionalities
- Only support esbuild for now - for webpack, we need Webpack 5 - a PR for WP5 support is here: https://github.com/jpmorganchase/modular/pull/1439


packages/modular-scripts/src/esbuild-scripts/plugins/rewriteDependenciesPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+      return acc;
+    },
+    {},
+  );
+
+  const dependencyRewritePlugin: esbuild.Plugin = {
+    name: 'dependency-rewrite',
+    setup(build) {
+      build.onResolve({ filter: /.*/, namespace: 'file' }, (args) => {
+        if (args.path in dependencies) {
```
LukeSheard, line 34: How do you handle subpath imports e.g. `react-dom/server` or `lodash/get`?

packages/modular-scripts/src/esbuild-scripts/config/createEsbuildConfig.ts
```diff
@@ -65,7 +67,7 @@ export default function createEsbuildConfig(
     color: !isCi,
     define,
     metafile: true,
-    tsconfig: paths.appTsConfig,
+    tsconfig: isApp ? paths.appTsConfig : undefined,
```
LukeSheard, line 70: I don't think we actually need tsconfig files for any app / view / etc and this is just a relic of CRA. We could probably remove this now?

packages/modular-scripts/src/esbuild-scripts/plugins/rewriteDependenciesPlugin.ts
cristiano-belloni: Good point, this needs refining.

packages/modular-scripts/src/build/index.ts
```diff
@@ -131,12 +151,13 @@ async function buildApp(target: string) {
...
 
   // Add dependencies from source and bundled dependencies to target package.json
-  const packageDependencies = await getPackageDependencies(target);
   const targetPackageJson = (await fs.readJSON(
     path.join(targetDirectory, 'package.json'),
   )) as CoreProperties;
   targetPackageJson.dependencies = packageDependencies;
-  targetPackageJson.bundledDependencies = Object.keys(packageDependencies);
+  if (isApp) {
+    targetPackageJson.bundledDependencies = dependencyNames;
```
cristiano-belloni, line 159: This needs to reflect the dependencies we actually bundle

packages/modular-scripts/src/esbuild-scripts/config/createEsbuildConfig.ts
cristiano-belloni: Done.

packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -210,74 +210,219 @@ describe('modular-scripts', () => {
...
-          "module": "dist-es/index.js",
+          "module": "static/js/index-IC6FL6E2.js",
           "name": "sample-view",
-          "typings": "dist-types/index.d.ts",
           "version": "1.0.0",
         }
       `);
     });
 
-    it('THEN outputs the correct output cjs file', () => {
```
LukeSheard, line 235: Why are removing the snapshots of the output? This is very useful for visualising bundle output changes when updating the configuration of builds.

packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -74,6 +90,8 @@ class DevServer {
     this.ws = ws(this.express);
 
     this.express.use(this.handleStaticAsset);
+    this.isApp || // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
LukeSheard, line 93: nit: this should be an if statement and would remove the eslint issues.

packages/modular-scripts/src/__tests__/index.test.ts
cristiano-belloni: We're removing the cjs test since now views build only esm modules.

packages/modular-scripts/src/esbuild-scripts/start/runtime/index.ts
```diff
@@ -107,3 +81,29 @@ connection.onmessage = (m: MessageEvent) => {
     isFirstCompilation = false;
   }
 };
+
```
LukeSheard, line 84: Why has this stuff moved?

packages/modular-scripts/src/esbuild-scripts/utils/openBrowser.ts
```diff
@@ -14,7 +14,7 @@ enum Actions {
   SCRIPT = 'Script',
 }
 
-const DEFAULT_BROWSER = process.env.BROWSER || OSX_CHROME;
+const DEFAULT_BROWSER = process.env.BROWSER || '';
```
LukeSheard, line 17: Comment here would be useful.

packages/modular-scripts/src/esbuild-scripts/start/index.ts
cristiano-belloni: It actually wouldn't - the problem is that `handleStaticAsset` returns a promise where a void function is expected, from the pov of the compiler.

packages/modular-scripts/src/utils/filterDependencies.ts
```diff
@@ -0,0 +1,37 @@
...
+  isApp: boolean,
+): FilteredDependencies {
+  if (isApp) {
+    return {
+      bundled: packageDependencies,
+      external: {},
+    };
+  }
+  const externalBlockList =
+    process.env.EXTERNAL_BLOCK_LIST && !isApp
```
LukeSheard, line 20: Why would we choose to block things? I think exposing this API without an understanding of why this is necessary is probably a bit careless.

packages/modular-scripts/src/esbuild-scripts/start/runtime/index.ts
cristiano-belloni: Lol, me fiddling with it and not putting it back. Doesn't make any difference but will move it again for better clarity.

packages/modular-scripts/src/start.ts
```diff
@@ -73,11 +65,16 @@ async function start(packageName: string): Promise<void> {
 
   // If you want to use webpack then we'll always use webpack. But if you've indicated
   // you want esbuild - then we'll switch you to the new fancy world.
-  if (!useWebpack || useEsbuild) {
+  if (!useWebpack || useEsbuild || isView) {
```
LukeSheard, line 68: This is untransparent for people that are still using webpack for apps they will probably be expecting webpack for starting a view. I think it would be better to throw an error down the webpack path if it's a view so that people are blocked from that flow, given esbuild is still opt in.

packages/modular-scripts/src/esbuild-scripts/utils/createViewTrampoline.ts
```diff
@@ -0,0 +1,66 @@
...
+</html>
+`;
+
+export async function createViewTrampoline(
+  fileName: string,
+  srcPath: string,
+  dependencies: Dependency,
+  browserTarget: string[],
+): Promise<esbuild.BuildResult & { outputFiles: esbuild.OutputFile[] }> {
+  const fileRelativePath = `./${fileName}`;
```
LukeSheard, line 21: Surely this would be the package name?

packages/modular-scripts/src/esbuild-scripts/start/runtime/index.ts
cristiano-belloni: Done

packages/modular-scripts/src/utils/filterDependencies.ts
cristiano-belloni: Mainly for three reasons:
1) If you have an organisation with a private registry, some internal packages could be present in the registry but not in the (possibly public) ESM CDN. In that case, you will need to bundle them.
2) You could be importing a dependency in the monorepo that "looks" like an external dep from the import statement but in reality is resolved by workspace symlinks (and it's not meant to be published to the registry or the CDN). In that case you almost always want to bundle it (we could filter out the workspace dependencies by default, but then you have the inverse problem: what if they're separately published to the CDN and you want to rewrite them?)
3) You could want an escape hatch for when a dependency dependency is broken in the CDN and you can't wait for the CDN to fix it. So you decide to be a bit less efficient and bundle it, until the CDN fixes it and you remove the blocklist and redeploy.

We can *also* have allow lists - they'd be useful in case the CDN exports only a selected range of dependencies (although less useful IMHO). But I wouldn't remove block lists.

cristiano-belloni: (or maybe you meant that `EXTERNAL_BLOCK_LIST` is a misnomer? It contains an optional list of dependencies that are to not be rewritten to an external CDN, but bundled in the build file. If you have a better name, happy to change it)

packages/modular-scripts/src/esbuild-scripts/utils/openBrowser.ts
cristiano-belloni: TL;DR: fixed with amore readable type assertion

This is interesting because the typescript version always failed in CI (it always tried to open Chrome the OSX way), but we never saw that because we didn't start a view with esbuild before.
I took a look, and it seems that the `OSX_CHROME` default was put there erroneously to shut up the compiler that failed later when trying to start a Node headless "browser". In reality, if we try to start Node we are sure that the argument is a string (because we test it), but the compiler is not smart enough to understand it. Giving it a false-ish default value works, because the Node path is never false-ish, while the browser path tests for false-ishness anyway. But you're right that' incomprehensible, so I fixed it with a type assertion. The "right-est" way to fix this would be probably rewriting it to use a discriminated union, but that will make the js and ts versions diverge, which is another problem not in scope here. IMHO we should aim, in the future, to use the ts version everywhere, but that's for another effort.

packages/modular-scripts/src/start.ts
cristiano-belloni: Agree. This is going to be temporary until we have Webpack 5, but I will log an error and throw down the webpack path.

packages/modular-scripts/src/esbuild-scripts/utils/createViewTrampoline.ts
cristiano-belloni: This is `index.js` if we're starting a view or `index-[HASH].js` if we're building. I think that the actual view entrypoint file should be hashed for cache-breaking purposes? After all whoever loads it fetches it at runtime, not downloads it from a registry and process it at build time, like the previous views. We keep track of the name in the output `package.json`, `module` field, so whoever loads it has a way to get the file name. If you think it should be named differently let's have a chat about that.

packages/modular-scripts/src/start.ts
cristiano-belloni: Done, in `start` and `build` too.

cristiano-belloni: There is [another PR](https://github.com/jpmorganchase/modular/pull/1439) that supports Webpack 5 and contains this PR. For the sake of simplicity, I'm keeping them separated.

cristiano-belloni:
Close this as it's contained in https://github.com/jpmorganchase/modular/pull/1439

PR #1391, opened 2022-02-09
Add modular serve <packageName>

LukeSheard:


packages/modular-scripts/src/program.ts
```diff
@@ -262,4 +262,17 @@ program
     await rename(oldPackageName, newPackageName);
   });
 
+interface ServeOptions {
+  port: string;
+}
+
+program
+  .command('serve <target>')
+  .description('Serves a pre-build modular app')
```
cristiano-belloni, line 271: pre-built

cristiano-belloni (approved):
lgtm

packages/modular-scripts/src/program.ts
LukeSheard: 👍

PR #1409, opened 2022-02-16
Various fixes

cristiano-belloni:
Various fixes, in preparation to https://github.com/jpmorganchase/modular/pull/1382

## Fix reload when serving an esbuild app / view

Various problems:

1. App not reloading when code changes -> the `isFirstCompilation` condition prevented the apps from reloading on change, fixed order.
1. Looking at the same app in >2 browser tabs caused an infinite loop of mutual reloading, because the ws server published to all apps when a new one connected. As a result, the "older" app would reload, find it's not their first build, re-connect. That would re-trigger a publish, cause the other one(s) to reload and so on. Fixed by publishing to all apps only on code change, but publishing only to the just-connected app on connection.
1. We were `watch`ing the build correctly, but not using the new build result anywhere. As a result, the browser tab would reload, but not show new content. Fixed by setting the new result with the build callback.

## Call `formatError` with the correct path

In esbuild-world, errors passed to `formatError` are relative to ModularRoot. If we pass them relative to the app path, instead, `formatError` tries to open a wrong path and kills the server with an exception. This fixes it.

## Stop Esbuild dev server from hanging

Sometimes the esbuild dev server hangs indefinitely when it's killed, either from terminal or programatically during tests. To solve the first issue,it seems it's sufficient to terminate its clients one by one. To solve the second issue, I'm sending a `kill -9` programatically.

## Revert react-error-overlay to previous patch and inject `process.env` and `process.platform`

It seems that in the latest patch, react-error-overlay is not only trying to access `process`, but the code it loads in the iframe it creates does too. In both cases, if `process` is not defined, react-error-overlay throws and is not available at runtime. For the first issue, the solution is to just inject `process` in the runtime. For the second issue, it's not clear how CRA does that (maybe at build time with a shim? I don't think they use the packaged react-error-overlay). Since this issue has been probably introduced by an automatic update, the solution is to inject process to solve issue 1 and revert to the previous patch to solve issue 2.


LukeSheard:
@cristiano-belloni Looks good! Can we create a changeset for each of the fixes? Then we can merge this!

packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -151,6 +156,12 @@ class DevServer {
       define: {
         global: 'window',
       },
+      banner: {
+        js: `window.process = {
```
LukeSheard, line 160: This probably needs a todo, web pack actually shims the process variable / other node.js concepts which is how they're defined. This is probably fine thought.

packages/modular-scripts/src/esbuild-scripts/start/plugins/wsReload.ts
```diff
@@ -13,43 +14,44 @@ export default function createPlugin(
...
+            },
+          }),
+        );
       };
 
-      server.on('connection', () => {
-        publish();
-      });
+      const publishAll = () => {
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
LukeSheard, line 40: Is there a way to resolve these lint warnings neatly?

LukeSheard: Maybe

```
return Promise.all(server.clients.map(publishClient));
```

packages/modular-scripts/src/esbuild-scripts/start/index.ts
cristiano-belloni: Yeah, that's the minimum viable injection to make react-error-overlay work.
There's a shim for it, yes (it even mocks complex stuff stuff like `hrtime` via the browser's timers) but our main problem with the newest version of reo is that the iframe tries to access `process`, and it's not trivial to inject it there.

packages/modular-scripts/src/esbuild-scripts/start/plugins/wsReload.ts
cristiano-belloni: That and a bunch of `void` typings did the trick.

LukeSheard (approved):

PR #1421, opened 2022-02-21
Webpack5 upgrade

cristiano-belloni:
Upgrade Webpack 4 -> Webpack 5.

- Configuration similarities with [CRA](https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpack.config.js) have been maintained where it makes sense.
- Output file names are slightly changed (see tests) because now, [for webpack 5, `chunkFileName` is only used for code splitting)](https://stackoverflow.com/questions/66077740/webpack-5-output-chunkfilename-not-working) (i.e. `chunkFileName` doesn't control vendor chunks; they're controlled by `filename` instead.
- Removed `optimize-css-assets-webpack-plugin` and introduced `css-minimizer-webpack-plugin` as advised in [optimize-css-assets-webpack-plugin's repo](https://github.com/NMFR/optimize-css-assets-webpack-plugin):
  > ⚠️ For webpack v5 or above please use [css-minimizer-webpack-plugin](https://github.com/webpack-contrib/css-minimizer-webpack-plugin) instead.


packages/modular-scripts/src/__tests__/app.test.ts
```diff
@@ -78,24 +78,26 @@ describe('when working with a NODE_ENV app', () => {
...
-            ├─ main.3db228f9.chunk.js.map #131rxqt
-            ├─ runtime-main.a0dc6a9b.js #o5bsr9
-            └─ runtime-main.a0dc6a9b.js.map #10n4p35"
+         ├─ js
+         │  ├─ main.6b9113da.js #1jh9hcj
+         │  ├─ main.6b9113da.js.map #2l1ifi
+         │  ├─ runtime-main.182069c4.js #r0hm4v
+         │  └─ runtime-main.182069c4.js.map #ul1xez
+         └─ media
+            └─ javascript,__webpack_public_path__ = __webpack_base_uri__ = htmlWebpackPluginPublicPath;.1feff74f.bin #1327v5u"
```
LukeSheard, line 96: What's this about?

cristiano-belloni: @LukeSheard - this is about Webpack 5 controlling the split chunks name via `fileName`, not `chunkFilename`, unlike Webpack 4 - `chunkFilename` is now only used for code splitting. See https://stackoverflow.com/a/67213952 (it's also mentioned in a a comment in this PR's webpack config).

cristiano-belloni: Ah, I see now you mean the `media` line. Wow, I've no idea, I'll investigate.

packages/modular-scripts/src/__tests__/app.test.ts
```diff
@@ -326,52 +326,53 @@ describe('when working with an app', () => {
...
+      │  ├─ App.css #1o0zosm
+      │  ├─ App.tsx #igvgtx
+      │  ├─ __tests__
+      │  │  └─ App.test.tsx #16urcos
+      │  ├─ index.css #o7sk21
+      │  ├─ index.tsx #zdn6mw
+      │  ├─ logo.svg #1okqmlj
+      │  └─ react-app-env.d.ts #t4ygcy
+      ├─ tsconfig.json #6rw46b
+      └─ tsconfig.tsbuildinfo #v0qc4l"
```
cristiano-belloni, line 348: This is probably another thing to remove

packages/modular-scripts/src/__tests__/app.test.ts
cristiano-belloni: @LukeSheard Fixed, this was the cause: https://github.com/jantimon/html-webpack-plugin/issues/1589

packages/modular-scripts/src/__tests__/app.test.ts
LukeSheard: Yeah, not sure why this is appearing? If it appears it should really be in the root? I wonder if this is todo with the type checking as part of the `fork-tsc-plugin`?

cristiano-belloni: Fixed

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -601,47 +557,60 @@ module.exports = function (webpackEnv) {
...
+                skipLibCheck: true,
+                inlineSourceMap: false,
+                declarationMap: false,
+                noEmit: true,
+              },
+            },
+            context: paths.appPath,
+            diagnosticOptions: {
+              syntactic: true,
+              semantic: true,
```
cangarugula, line 589: We're hard coding tsconfig here instead of providing a path to the tsconfig file. Why are we locking this down? Are there specific settings needed for this upgrade?

packages/modular-scripts/src/__tests__/app.test.ts
```diff
@@ -197,7 +197,7 @@ describe('When working with a nested app', () => {
     ).toMatchSnapshot();
   });
 
-  it('can generate a css/main.a0f92c83.chunk.css', async () => {
+  it('can generate a css/main.1a7488ce.css', async () => {
```
cangarugula, line 200: This and subsequent tests have the file name hard coded in the tests - can we make this dynamic?

packages/modular-scripts/src/__tests__/app.test.ts
```diff
@@ -326,52 +326,52 @@ describe('when working with an app', () => {
...
+      └─ tsconfig.json #6rw46b"
+    `);
   });
 
   it('can build an app', () => {
     expect(tree(path.join(modularRoot, 'dist', 'sample-app')))
       .toMatchInlineSnapshot(`
       "sample-app
-      ├─ asset-manifest.json #131f6kc
+      ├─ asset-manifest.json #nzl5za
```
cangarugula, line 355: I think it'd be good to have a snapshot of the contents of the asset-manifest.json and index.html so we can see explicit how build output changes

packages/modular-scripts/src/__tests__/utils/createEsbuildBrowserslistTarget.test.ts
```diff
@@ -51,7 +51,7 @@ describe('createEsbuildBrowserslistTarget', () => {
     );
 
     const result = createEsbuildBrowserslistTarget(process.cwd());
-    expect(result).toEqual(['chrome97', 'firefox96', 'safari15.2']);
+    expect(result).toEqual(['chrome99', 'firefox97', 'safari15.2']);
```
cangarugula, line 54: Can we save the list of unsupported browsers somewhere in the app? It'll help us ensure we are consistent with which browsers we can filtering out and also it'll prevent these hard coded tests from breaking if we miss a line when the list needs to be updated?

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: Just in the spirit of remaining compatible with CRA. But I'm quite neutral on this: if we think the overrides should be removed, no problem.

packages/modular-scripts/src/__tests__/app.test.ts
cristiano-belloni: Agree that we should make them decoupled from the hashes (for example: "it can generate a style hashed main chunk in the css directory"). Not sure if you mean that with dynamic?

packages/modular-scripts/src/__tests__/app.test.ts
cristiano-belloni: We already have a snapshot of both, but it's not inline. Do you propose an inline snapshot to see it in the test code?

packages/modular-scripts/src/__tests__/utils/createEsbuildBrowserslistTarget.test.ts
LukeSheard: Yeah - these tests are separate to any web pack piece of this. It just shows that our current version of browsers list is out of date and it's been updated as part of this PR. We could tackle this part of the changeset by updating outside of this context.

packages/modular-scripts/src/__tests__/app.test.ts
steveukx: New issue created: #1481

steveukx (approved):

PR #1433, opened 2022-02-24
fix: package builds respect --private option

benpryke:


LukeSheard:
Created a changeset, will merge once checks have completed again.

LukeSheard (approved):

LukeSheard:
@benpryke Although, just seen there's no test to cover this case. Can we add one?

cristiano-belloni (approved):
LGTM (and I agree with https://github.com/jpmorganchase/modular/pull/1433#issuecomment-1050886438)

cristiano-belloni:
> Is some additional setup required? I couldn't see any docs to suggest that. If this is actually broken right now, I'd rather not fix the world for a one-line patch, but I don't disagree that there's a gap for a couple of cross-package build tests.

Hi Ben, unfortunately my laptop is out of order atm and I'm waiting for it to be fixed so I can't try, but for me no additional step was required to run the tests, apart from the usual `yarn` and `yarn build` at the root level. Github Actions run the whole test suite on each PR commit and merging is disallowed if the tests are not passing, so it's unlikely that they are broken (although there might be a bug that's triggered only in your system).

Could you try to tun them with `CI=true`? Could you also take a look at what is breaking exactly? I know, for example, that tests will fail if the repository has an uncommitted change, because of the way we wrap "atomic" commands on our test packages. It has bitten me more than once when running `yarn` results in a modification of `yarn.lock`, for example.

LukeSheard:
@benpryke make sure you're running `yarn test` - I've raised a PR to clarify reasoning for this.

packages/modular-scripts/src/build/buildPackage/makeBundle.ts
```diff
@@ -183,7 +183,10 @@ export async function makeBundle(
           // Let's collect the name and add it in the package.json
           // we publish to the registry
           // TODO: make sure local workspaces are NOT explicitly included in package.json
-          if (packageJsons[importedPackage].private !== true) {
+          if (
+            includePrivate ||
+            packageJsons[importedPackage].private !== true
+          ) {
             localImports[importedPackage] = packageJsons[importedPackage]
```
LukeSheard, line 190: Maybe let's add a console warning here if we are in the case where we includePrivate and also hit this branch?

LukeSheard:
LGTM apart from my one comment.

packages/modular-scripts/src/build/buildPackage/makeBundle.ts
benpryke: I think it's perfectly reasonable to break projects down into multiple packages that it wouldn't make sense to publish on their own, so I don't think a warning is warranted every time someone builds a project that relies on unpublished packages in this way.

LukeSheard: The use case of building a package for publishing implies that all the dependencies are being published and maybe worth considering if a packages should be public or not. I wouldn’t expect anyone to be using the private flag for building on packages which aren’t being published. 

Given that private package building is opt in I would put a warning to show that the package being built is depending on other pieces which would need to be build opt in.

benpryke: I think there's a distinction between publishing the source and building a distribution. The private field is simply a way of specifying that a package should not be published standalone and implies nothing about the build process or how it's otherwise used.

I presume users will most commonly add the private flag after encountering a build failure where they used private package dependencies in their project.

Given the above, it still strikes me as odd to repeatedly display a warning for something the user has opted into.

steveukx: I'd suggest we merge as-is and split out a new feature of debug logging that would be useful for notice kind of messages like this.

PR #1439, opened 2022-02-28
Build ESM views with external CDN support - esbuild & webpack 5

cristiano-belloni:
This PR combines:

- https://github.com/jpmorganchase/modular/pull/1421
- https://github.com/jpmorganchase/modular/pull/1382

to support ESM views built with Webpack 5. In other words, it makes it possible to build a view with ESM CDN support using Webpack 5.

### TODO

- [x] Build an ESM view with dependency rewriting and trampoline
- [x] Support multi-file imports rewriting in Webpack 5 - see https://webpack.js.org/guides/author-libraries#external-limitations
- [x] Support local CSS in standalone mode
- [x] Start an ESM view with dependency rewriting and trampoline
- [x] Tests


LukeSheard:
@cristiano-belloni could we separate this from ESM support? This looks like a good feature standalone which we could merge.

cristiano-belloni:
> @cristiano-belloni could we separate this from ESM support? This looks like a good feature standalone which we could merge.

We already have standalone Webpack 5 support in https://github.com/jpmorganchase/modular/pull/1421 - or are you referring to something else in the PR?

PR #1447, opened 2022-03-02
Remove init commit after create-modular-react-app runs

cangarugula:
Removing the git commit message of "Initial Commit" after project has been scaffolded. This still stages the files, but doesn't commit them to give the developer greater control over their work.


cristiano-belloni (approved):
lgtm

PR #1486, opened 2022-03-16
Enable pre-release workflow

cristiano-belloni:
### Workflow proposal

1. Every time we want to make a pre-release, we create a  `release/...` long-lived branch
2. In the release branch, we enter pre mode. Normally we want to set the npm tag to be `next` (`yarn changeset pre enter next`), but if we want to have concurrent tags, we can have them (`yarn changeset pre enter other_tag`)
3. We branch off the release branch and create PRs to it as it was `main`, updating our changesets normally
4. Like with main, every time we push to the release branch, a changeset-release branch is created (for example, `release/webpack-5` will generate `changeset-release/release/webpack-5`). If that changeset-release branch is merged back into the release branch, a prerelease version should be published to the npm tag we chose at point 2 (normally `next`)
5. When the release branch is finished and ready to go into main, we exit pre mode (`yarn changeset pre exit`) and we create a PR from the release branch into `main`. If we forget to exit pre mode, the CI will catch it and not let us merge back into `main`

Summary of the upcoming webpack 5 pre-release after discussing it, for reference (thanks @steveukx):

- As described in this PR we will have a release/webpack-5 to contain the upgrade from webpack 4 -> 5
- Publish the changes to @alpha-webpack5 tag on npm
- Once testing is complete, would mean merging back into main and an @alpha-webpack5 tag pushed to npm
- Any change in main will need to be merged into @alpha-webpack5 (process to be owned by the developer)
- On completing testing, deprecate or delete the tagged version 

### Changes in this PR

Preparatory work for the [changeset pre-release](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) [flow](https://github.com/changesets/action/issues/69#issuecomment-774909280)

- Allow `release/*` branches to trigger the release workflow
- Allow `release/*` branches to trigger tests
- Allow `release/*` branches to trigger static analysis
- Allow `release/*` branches to trigger validate
- Fail checks if any PR to main (or commit to main itself) is in pre-release mode
- Upgrade action changeset to `@v1` ([master is deprecated](https://github.com/changesets/action/tree/master#changesets-release-action))


LukeSheard:
I'm not sure this will have the desired behaviour, since running the pre-release check on any CI including release will also fail. 

I would suggest that if we want to setup a pre-release branch we setup CI so that a dedicated branch creates a prerelease PR on that base branch if we merge into there. Then we can enable / disable this workflow depending on whether we want to use it or not.

cristiano-belloni:
> since running the pre-release check on any CI including release will also fail.

Yeah, that's true. I can duplicate the workflow for main / release branches though, I'd like to avoid having a dedicated branch and turning on and off stuff because it's easy to get it wrong.

cristiano-belloni:
@LukeSheard I created a workflow to check pre mode, it's a lightweight (3 secs) concurrent check. I also added the desired workflow in the description, if it makes sense, so we can comment on it.

benpryke:
Should you also run the validate workflow on `release/**`?

LukeSheard:
In line with this I've also disabled direct pushes to `release/**` branches so this makes sense.

cristiano-belloni:
> In line with this I've also disabled direct pushes to `release/**` branches so this makes sense.

Thanks!

> Would it not also make sense to create a check which merges upstream changes from `main` into `release/**` branches? Or at least creates a PR for it?

I don't think we want it automatic right now because when we merge upstream we want to regenerate the `pre.json` too, and I'd like to handle that manually, at least for the first times.

benpryke (approved):

PR #1503, opened 2022-03-24
Relax our dependency criteria

cristiano-belloni:
- Allow devDependencies to be in the dependency manifest
- Prevent throwing when generating the manifest if there are transitive deps, show a warning that this will be an error in next major

### DIFF
--- a/packages/modular-scripts/src/utils/getPackageDependencies.ts
+++ b/packages/modular-scripts/src/utils/getPackageDependencies.ts
@@ -74,7 +74,7 @@ export async function getPackageDependencies(
         rootPackageJsonDevDependencies[depName];
       if (!depVersion) {
         logger.error(
-          `Package ${depName} imported in ${target} source but not found in package dependencies or hoisted dependencies`,
+          `Package ${depName} imported in ${target} source but not found in package dependencies or hoisted dependencies - this will prevent you from continuing in the next release of modular`,
         );
       }
       manifest[depName] = depVersion;
--- a/packages/modular-scripts/src/utils/getPackageDependencies.ts
+++ b/packages/modular-scripts/src/utils/getPackageDependencies.ts
@@ -5,6 +5,7 @@ import type { CoreProperties } from '@schemastore/package';
 import getModularRoot from './getModularRoot';
 import getLocation from './getLocation';
 import getWorkspaceInfo from './getWorkspaceInfo';
+import * as logger from './logger';
 
 type DependencyManifest = NonNullable<CoreProperties['dependencies']>;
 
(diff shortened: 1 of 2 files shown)


packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -46,19 +47,18 @@ export async function getPackageDependencies(
...
-      ) as CoreProperties
-    ).dependencies || {};
+  const targetManifest = fs.readJSONSync(
+    path.join(targetLocation, 'package.json'),
+  ) as CoreProperties;
+
+  const rootPackageJsonDependencies = rootManifest.dependencies || {};
+  const targetPackageJsonDependencies = targetManifest.dependencies || {};
+  const rootPackageJsonDevDependencies = rootManifest.devDependencies || {};
+  const targetPackageJsonDevDependencies = targetManifest.devDependencies || {};
```
steveukx, line 61: nit: as an option for making this a bit less verbose:

```javascript
const deps = Object.assign(Object.create(null).
  targetManifest.devDependencies,
  rootManifest.devDependencies, 
  targetManifest.dependencies, 
  rootManifest.dependencies, 
);

// ...
const devVersion = deps[depName];
if (depVersion === undefined) {
  logger.error(...);
}

manifest[depName]
```

steveukx:
Are there any tests of the `build` command that could cover this change?

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Done with 

```
const deps = {
    ...rootManifest.dependencies,
    ...targetManifest.dependencies,
    ...rootManifest.devDependencies,
    ...targetManifest.devDependencies,
  };
```

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -55,10 +55,12 @@ export async function getPackageDependencies(
     path.join(targetLocation, 'package.json'),
   ) as CoreProperties;
 
-  const rootPackageJsonDependencies = rootManifest.dependencies || {};
-  const targetPackageJsonDependencies = targetManifest.dependencies || {};
-  const rootPackageJsonDevDependencies = rootManifest.devDependencies || {};
-  const targetPackageJsonDevDependencies = targetManifest.devDependencies || {};
+  const deps = {
+    ...rootManifest.dependencies,
```
steveukx, line 59: Are `.dependencies` and `.devDependencies` always objects? If they could be undefined, this will throw.

cristiano-belloni: They're either `undefined` or object. I don' think they will throw though:

```
a = undefined
b = { c: 'hey' }
{ ...a, ...b }
{c: 'hey'}
```

steveukx: ^-- ignore me, non-iterable only throws when spreading into an array.

cristiano-belloni: I'll resolve this

steveukx (approved):

PR #1517, opened 2022-03-28
fix node-12 incompatible javascript

LukeSheard:


babel.config.js
```diff
@@ -12,17 +12,13 @@ module.exports = (api) => {
         '@babel/preset-env',
         {
           targets: {
-            node: 'current',
+            // TODO: can we get this to read from package.json somehow..?
```
benpryke, line 15: Can you not read `package.json` and parse the `engines.node` field to find the smallest number?

steveukx (approved):

cristiano-belloni (approved):

babel.config.js
```diff
@@ -12,17 +12,13 @@ module.exports = (api) => {
...
-      '@babel/preset-react',
-    ],
-    overrides: [
-      {
-        test: /\.tsx?$/,
-        presets: ['@babel/preset-typescript'],
-      },
+      '@babel/preset-typescript',
     ],
+    plugins: ['@babel/plugin-proposal-class-properties'],
```
benpryke, line 22: Q: why do we need to add this additional plugin on top of `@babel/preset-typescript`? Are there any more we need?

PR #1538, opened 2022-04-05
Drop support for node 12

cristiano-belloni:
Remove support for node 12, as it's reaching end of of maintenance at the end of April 2022 - https://nodejs.org/en/about/releases/

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -12,13 +12,17 @@ module.exports = (api) => {
         '@babel/preset-env',
         {
           targets: {
-            // TODO: can we get this to read from package.json somehow..?
-            node: '12',
+            node: 'current',
           },
         },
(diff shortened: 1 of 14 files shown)


babel.config.js
```diff
@@ -12,13 +12,17 @@ module.exports = (api) => {
         '@babel/preset-env',
         {
           targets: {
-            // TODO: can we get this to read from package.json somehow..?
-            node: '12',
+            node: 'current',
```
LukeSheard, line 15: This is subjective. We should hard code this otherwise it's inherited from whatever workflow node version is running which can make updating the `setup-node` flow breaking.

cristiano-belloni: I hardcoded node14.

PR #1546, opened 2022-04-06
support node 16 explicitly and run test workflow against target node

LukeSheard:
Updating the `setup-node` configuration so that the versions of node that we support are explictly coded in, this is because currently we're running a workflow with the latest node `v.x` version in a range, not the minimum version we support which leads to runtime differences.

cc. @steveukx / @cristiano-belloni


packages/modular-scripts/package.json
```diff
@@ -13,7 +13,7 @@
     "./tsconfig.json": "./tsconfig.json"
   },
   "engines": {
-    "node": "^12.13.0 || ^14.15.0 || >=15.0.0"
+    "node": "^12.20.0 || ^14.15.0 || >=15.0.0 || >=16.0.0"
```
cristiano-belloni, line 16: Since we're updating the engines, wouldn't it be better to merge this in https://github.com/jpmorganchase/modular/pull/1538 and have teams testing it in alpha for a while? I bet that some teams run node < 12.20, for example.

LukeSheard: If you look at the original commit here, installation fails for `<12.20.0` unless `--ignore-engines` is already set because of the current `find-up` dependency. This was actually one of the reasons that I'm proposing this change to ensure that dependency upgrades also don't cause failures in the version of node we support. 

In theory this release should be a no-op in a practical sense.

PR #1561, opened 2022-04-11
Dependency filtering improvements: allow list + wildcards

cristiano-belloni:
We need a way to control which dependencies get bundled to an ESM view and which dependencies get rewritten to ESM CDN. Before we had only a block list to exclude dependencies from rewriting (ie bundle them). I had feedback from @susanafcosta that she would like to have the inverse mechanism too - an allow list to specify dependencies to be rewritten and that she would like to specify multiple dependencies via wildcard instead of enumerating all of them one by one (for example to include / exclude an entire scope)

This is how it's implemented in this PR. I'm not sure that block and allow list are the correct terms at this point terms - if you have proposals for better terms I'm ok to change them:

- `EXTERNAL_ALLOW_LIST` is a comma-separated string that specifies which dependencies are allowed to be rewritten to CDN; if not specified, its default value is `'*'` ( -> all dependencies are rewritten)
- `EXTERNAL_BLOCK_LIST` is a comma-separated string that specifies which dependencies are **not** allowed to be rewritten to CDN; if not specified its default value is `''` ( -> no dependency excluded, i.e. all dependencies are rewritten)
- If a dependency is in both lists, it will **not** be rewritten (`EXTERNAL_BLOCK_LIST` wins)
- If a dependency is in none of the lists, it will **not** be rewritten (but remember that all dependencies are allowed if one list is not specified)
- Wildcards / glob patterns [are supported](https://github.com/micromatch/micromatch) in both lists


packages/modular-scripts/src/__tests__/index.test.ts
```diff
@@ -421,16 +421,134 @@ describe('modular-scripts', () => {
...
+    it('THEN rewrites only the dependencies that are not specified in the blocklist', async () => {
+      const baseDir = path.join(
+        modularRoot,
+        'dist',
+        'sample-view',
+        'static',
+        'js',
+      );
+
+      const indexFile = (
```
LukeSheard, line 472: I would just snapshot this so that we can assert the import syntax renames the same across esbuild upgrades

packages/modular-scripts/src/utils/filterDependencies.ts
```diff
@@ -16,16 +17,29 @@ export function filterDependencies(
...
+
+  // By default, nothing is in blocklist
   const externalBlockList =
     process.env.EXTERNAL_BLOCK_LIST && !isApp
       ? process.env.EXTERNAL_BLOCK_LIST.split(',')
       : [];
+
+  // By default, everything in allow list
+  const externalAllowList =
+    process.env.EXTERNAL_ALLOW_LIST && !isApp
```
LukeSheard, line 29: We should add documentation for this before we get too deep adding APIs and can't remember what's there and not.

packages/modular-scripts/src/__tests__/index.test.ts
cristiano-belloni: Done, snapshotting all the index files.

LukeSheard: Probably best to use the format we have elsewhere where the file is run through prettier so that the snapshot is readable. See [here](https://github.com/jpmorganchase/modular/blob/main/packages/modular-scripts/src/__tests__/app.node-env.test.ts#L136) for an example from one of the other build tests

cristiano-belloni: Good catch, thanks. Done!

PR #1651, opened 2022-04-28
Remove CHANGELOG.md files from templates

LukeSheard:
When initializing from a default template we copy the `CHANGELOG.md` file from the template folder into the target directory which is a little bit silly. These files are included by default in `npm-packlist` so we have to manually exclude them when unpacking.

### DIFF
--- a/.changeset/thirty-dolphins-impress.md
+++ b/.changeset/thirty-dolphins-impress.md
@@ -1,0 +1,6 @@
+---
+'modular-scripts': patch
+---
+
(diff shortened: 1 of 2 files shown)


cristiano-belloni:
https://github.com/jpmorganchase/modular/pull/1636/files#diff-6429cfff0b0161203f9bc77f5d10826e933a3c790bcc286769532009889ee269R151 -> We already filter out extraneous files by globbing package.json's `[file]`s. This should remove any extraneous file, including changelog.md

PR #1693, opened 2022-05-06
Rewrite pinned dependencies from lockfile

cristiano-belloni:


packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -735,20 +742,28 @@ module.exports = function (webpackEnv) {
...
+    }
+    return {
       ...acc,
       [name]: externalCdnTemplate
         .replace('[name]', name)
-        .replace('[version]', version),
-    }),
-    {},
-  );
+        .replace('[version]', version || externalResolutions[name])
```
LukeSheard, line 763: Is there a case where version would not be defined?

packages/modular-scripts/src/esbuild-scripts/api.ts
```diff
@@ -50,10 +61,16 @@ ReactDOM.render(<Component />, DOMRoot);`;
...
         },
       },
-      createRewriteDependenciesPlugin({
-        ...dependencies,
-        'react-dom': dependencies.react,
-      }),
+      createRewriteDependenciesPlugin(
+        {
+          ...dependencies,
+          'react-dom': dependencies.react,
```
LukeSheard, line 67: Surely `react-dom` must be a dependency in the tree for starting a view to work, so we should use the resolution version from the repo in the case where `react-dom` is a different version to `react`?

packages/modular-scripts/src/esbuild-scripts/api.ts
```diff
@@ -11,10 +11,21 @@ import type { Dependency } from '@schemastore/package';
 
 type FileType = '.css' | '.js';
 
+export const indexFile = `
```
LukeSheard, line 14: Seems a bit non-descript. Is this not `DEFAULT_ESM_VIEW_INDEX_TEMPLATE`?

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: > Is there a case where version would not be defined?

Good question. There is at the moment, because we relaxed our criteria to only warn in case a version is nor present in (on e of the) package.json. The reason is that some widely used libraries rely on transitive dependencies (dm me for more info)

packages/modular-scripts/src/esbuild-scripts/api.ts
cristiano-belloni: Done.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -55,30 +56,67 @@ export async function getPackageDependencies(
...
+    targetManifest.devDependencies,
+    targetManifest.dependencies,
   ) as Dependency;
 
-  /* Get regular dependencies from package.json (regular) or root package.json (hoisted)
-   * Exclude workspace dependencies. Error if a dependency is imported in the source code
+  const lockDeps = await buildDepTree(
+    // Build a dependency tree from the lockfile, using target dependencies and root dependencies in order of specificity
+    JSON.stringify(
+      Object.assign(Object.create(null), targetManifest, {
```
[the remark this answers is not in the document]
cristiano-belloni: This is following a comment by @steveukx - there might be dependencies called as object default members.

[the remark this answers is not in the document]
steveukx: This one would be fine as a `{ ...spread }` - it wasn't in the other case because the resulting object was used with a `if (key in object)` check without an `object.hasOwnProperty(key)` filter.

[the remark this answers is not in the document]
steveukx: This one would be fine as a `{ ...spread }` - it wasn't in the other case because the resulting object was used with a `if (key in object)` check without an `object.hasOwnProperty(key)` filter.

steveukx (approved):

PR #1707, opened 2022-05-10
Move to Webpack 5 Asset Modules

cheemcheem:
Webpack 5 in modular-scripts@3.0.0 doesn’t compile fonts correctly, and instead you end up with font files in the root directory that have JavaScript inside of them (this does not work in the browser).
 
I did some research and found that Webpack 5 recommends a new way of loading assets which is using asset modules instead of the various url/file/raw loaders.

Webpack 5 asset modules docs: https://webpack.js.org/guides/asset-modules/
StackOverflow answer that pointed me in this direction: https://stackoverflow.com/a/68913794


pedroBlazquez:
+1 to support  this change. We are seeing the same issue in our build

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -53,10 +53,6 @@ const reactRefreshOverlayEntry = require.resolve(
 // makes for a smoother build process.
 const shouldInlineRuntimeChunk = process.env.INLINE_RUNTIME_CHUNK !== 'false';
 
-const imageInlineSizeLimit = parseInt(
-  process.env.IMAGE_INLINE_SIZE_LIMIT || '10000',
-);
-
```
cheemcheem, line 59: Limit was no longer necessary since the 'asset/resource' type is never inlined / is always outputted to a separate file. 

If we want some to be inline and others to be outputted in their own file, we can change the type 'asset' and according to the [docs](https://webpack.js.org/guides/asset-modules/#general-asset-type):

> Now webpack will automatically choose between resource and inline by following a default condition: a file with size less than 8kb will be treated as a inline module type and resource module type otherwise.

(Note: 'resource' means outputted to its own file in this context.)

packages/modular-scripts/src/__tests__/app.test.ts
```diff
@@ -143,14 +143,16 @@ describe('When working with a nested app', () => {
...
+         ├─ js
+         │  ├─ 788.bbd34b33.js #33pg04
+         │  ├─ 788.bbd34b33.js.LICENSE.txt #eplx8h
+         │  ├─ 788.bbd34b33.js.map #1xdy7n0
+         │  ├─ main.f646b520.js #15ws1rq
+         │  ├─ main.f646b520.js.map #b2tlah
+         │  ├─ runtime-main.de012fdc.js #1qz643h
+         │  └─ runtime-main.de012fdc.js.map #ntuwq4
+         └─ media
+            └─ logo.103b5fa1..svg #1okqmlj"
```
cheemcheem, line 155: This logo.svg file being exported on its own is probably a result of this PR using the 'asset/resource' type.

If we instead use the 'asset' type, it would set an inline size limit of 8kb as I mentioned in my above comment and this logo file probably wouldn't be exported on its own.

cheemcheem: Resolved by moving to 'asset' type.

packages/modular-scripts/react-scripts/config/webpack.config.js
cheemcheem: Moved to 'asset' type which mimics this functionality but with an 8kb limit instead of 10kb.

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -360,35 +356,15 @@ module.exports = function (webpackEnv) {
...
-            // was handling them
-            {
-              test: /\.svg$/,
-              use: [
-                require.resolve('@svgr/webpack'),
-                require.resolve('url-loader'),
-              ],
-            },
             // TODO: Merge this config once `image/avif` is in the mime-db
             // https://github.com/jshttp/mime-db
```
cheemcheem, line 360: I'm quite certain that image/avif is in the mime-db in https://github.com/jshttp/mime-db, but didn't want to go out of scope and remove this line in this PR.

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -360,35 +356,15 @@ module.exports = function (webpackEnv) {
...
-            // smaller than specified limit in bytes as data URLs to avoid requests.
-            // A missing `test` is equivalent to a match.
-            {
-              test: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
-              loader: require.resolve('url-loader'),
-              options: {
-                limit: imageInlineSizeLimit,
-                name: 'static/media/[name].[hash:8].[ext]',
+              test: /\.(bmp|gif|jpe?g|png|svg|avif|ico|eot|ttf|woff2?)(\?v=\d+\.\d+\.\d+)?$/i,
+              type: 'asset',
```
cheemcheem, line 365: Using type 'asset' so that WebPack will determine whether to inline an asset using an inline file size limit of 8kb (I couldn't see a way to customise the 8kb limit back to the 10kb limit that was here, but they probably chose 8kb for best practice reasons so I'm not too concerned about that).

cristiano-belloni:
@cheemcheem - I extended your PR to use the same configuration for SVG and AVIF as [CRA5](https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpack.config.js) - we strive to be compatible as we have commands that port CRA apps to a Modular repo.

cheemcheem:
Changes work as expected on my side 👍

PR #1711, opened 2022-05-11
Add the analyze command

cristiano-belloni:
```
> modular-dev analyze my-package
{
  "manifest": {
    "react": "^16.8.1"
  },
  "resolutions": {
    "react": "16.14.0"
  }
}
```

### DIFF
--- a/packages/modular-scripts/src/analyze.ts
+++ b/packages/modular-scripts/src/analyze.ts
@@ -1,0 +1,8 @@
+import actionPreflightCheck from './utils/actionPreflightCheck';
+import { getPackageDependencies } from './utils/getPackageDependencies';
+
+async function analyze({ target }: { target: string }): Promise<void> {
+  console.log(JSON.stringify(await getPackageDependencies(target), null, 2));
+}
+
+export default actionPreflightCheck(analyze);
--- a/packages/modular-scripts/src/program.ts
+++ b/packages/modular-scripts/src/program.ts
@@ -59,6 +59,16 @@ program
   );
 
 program
+  .command('analyze <package-name>')
+  .description(`Analyze the dependencies of a package.`)
+  .action(async (packageName: string) => {
+    const { default: analyze } = await import('./analyze');
+    return analyze({
+      target: packageName,
+    });
+  });
+
+program
   .command('build <packages...>')
   .description(
     'Build a list of packages (multiple package names can be supplied separated by space)',


docs/commands/analyze.md
```diff
@@ -0,0 +1,29 @@
...
+The source code of the specified package is statically analyzed to find all the
+`import` or `require` statements. Then, for each external dependency imported in
+the code, its `manifest` version is computed by looking at the workspace's
+manifest files, and its `resolution` version is computed by parsing the
+workspace's `yarn.lock`.
+
+For this reason, `resolution` versions are strictly exact, while `manifest`
+versions can contain any ammissible type of
+[ranged semver](https://github.com/npm/node-semver#versions). `manifest` version
+are computed by returning after the first successful lookup between these
```
benpryke, line 23: Would this be clearer: "`manifest` versions are set to the first successful lookup from the following sequence of lookup locations:"?

docs/commands/analyze.md
```diff
@@ -0,0 +1,29 @@
...
+
+The source code of the specified package is statically analyzed to find all the
+`import` or `require` statements. Then, for each external dependency imported in
+the code, its `manifest` version is computed by looking at the workspace's
+manifest files, and its `resolution` version is computed by parsing the
+workspace's `yarn.lock`.
+
+For this reason, `resolution` versions are strictly exact, while `manifest`
+versions can contain any ammissible type of
+[ranged semver](https://github.com/npm/node-semver#versions). `manifest` version
```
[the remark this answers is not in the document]
cristiano-belloni: Thx, done!

docs/commands/analyze.md
```diff
@@ -0,0 +1,29 @@
...
+
+Analyzes the dependencies of a package from its source code, emitting JSON to
+`stdout` (please note that errors and warnings can still be emitted to
+`stderr`).
+
+## How it works
+
+The source code of the specified package is statically analyzed to find all the
+`import` or `require` statements. Then, for each external dependency imported in
+the code, its `manifest` version is computed by looking at the workspace's
```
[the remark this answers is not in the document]
cristiano-belloni: Added a section with description of the output + example

docs/commands/analyze.md
cristiano-belloni: It would. Done.

docs/commands/analyze.md
```diff
@@ -0,0 +1,54 @@
...
+## How versions are calculated
+
+The source code of the specified package is statically analyzed to find all the
+`import` or `require` statements. Then, for each external dependency imported in
+the code, a its `manifest` version is computed by looking at the workspace's
+manifest files, and its `resolution` version is computed by parsing the
+workspace's `yarn.lock`.
+
+For this reason, `resolution` versions are strictly exact, while `manifest`
+versions can contain any ammissible type of
```
steveukx, line 46:
```suggestion
versions can contain any admissible type of
```

packages/modular-scripts/src/analyze.ts
```diff
@@ -0,0 +1,8 @@
+import actionPreflightCheck from './utils/actionPreflightCheck';
+import { getPackageDependencies } from './utils/getPackageDependencies';
+
+async function analyze({ target }: { target: string }): Promise<void> {
+  console.log(JSON.stringify(await getPackageDependencies(target), null, 2));
```
cangarugula, line 5: Can we wrap this with `strip-ansi` to strip escape characters? This will be important for cross platform users

docs/commands/analyze.md
```diff
@@ -0,0 +1,54 @@
...
+[ranged semver](https://github.com/npm/node-semver#versions). `manifest`
+versions are set to the first successful lookup from the following sequence of
+lookup locations::
+
+1. The package's `package.json` `dependencies` field
+1. The package's `package.json` `devDependencies` field
+1. The root `package.json` `dependencies` field
+1. The root `package.json` `devDependencies` field
```
cangarugula, line 54: This numbered list only has `1.`

docs/commands/analyze.md
```diff
@@ -0,0 +1,54 @@
...
+`import` or `require` statements. Then, for each external dependency imported in
+the code, a its `manifest` version is computed by looking at the workspace's
+manifest files, and its `resolution` version is computed by parsing the
+workspace's `yarn.lock`.
+
+For this reason, `resolution` versions are strictly exact, while `manifest`
+versions can contain any admissible type of
+[ranged semver](https://github.com/npm/node-semver#versions). `manifest`
+versions are set to the first successful lookup from the following sequence of
+lookup locations::
```
steveukx, line 49: What will be the result if there is no successful lookup?

cristiano-belloni: Right now, we will warn on stderr (so that the user can still capture stdout, without the match). In a future release, we will be stricter and abort the command with err return value.

docs/commands/analyze.md
cangarugula: I thought this section was to specify the order in which you perform the look up for dependency versions. Wouldn't it make sense to make it an ordered list so it's clear the order of logic?

.changeset/lemon-rules-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"modular-scripts": minor
+---
+
+Add the analyze command
```
cangarugula, line 5: 'Add the `modular analyze` command'?

docs/commands/analyze.md
cristiano-belloni: > Wouldn't it make sense to make it an ordered list so it's clear the order of logic?

This is an ordered list, and its order is implicit in the sequential ordering in the text. Users will see it rendered with the "correct" numbers. If you use the same number, it's just easier (and less prone to errors) to add more steps in the middle with just one line change in the future. That's why I normally use 1. everywhere (unless I have to reference the lines by number in the text).

Anyway, it makes little difference :) If you feel explicit numbers are better, I'll change them.

cristiano-belloni: Done!

.changeset/lemon-rules-kick.md
cristiano-belloni: Better - done.

docs/commands/analyze.md
cangarugula: Ahhh I see I see. This is a markdown implementation. You can ignore :)

cristiano-belloni: > Ahhh I see I see. This is a markdown implementation. You can ignore :)

I just changed them 🤣

PR #1719, opened 2022-05-13
Use PUBLIC_URL for dist manifest

cristiano-belloni:


packages/modular-scripts/src/build/index.ts
```diff
@@ -230,8 +230,8 @@ async function buildStandalone(
       modular: targetPackageJson.modular,
       dependencies: targetPackageJson.dependencies,
       bundledDependencies: targetPackageJson.bundledDependencies,
-      module: jsEntryPoint,
-      style: cssEntryPoint,
+      module: paths.publicUrlOrPath + jsEntryPoint,
```
cangarugula, line 233: I think you'll need to do path.join to resolve the path. If I pass in "./" as the PUBLIC_URL, it would resolve to ".//static/...etc" with a double slash.

cristiano-belloni: > I think you'll need to do path.join to resolve the path. If I pass in "./" as the PUBLIC_URL, it would resolve to ".//static/...etc" with a double slash.

done!

packages/modular-scripts/src/build/index.ts
```diff
@@ -230,8 +230,12 @@ async function buildStandalone(
       modular: targetPackageJson.modular,
       dependencies: targetPackageJson.dependencies,
       bundledDependencies: targetPackageJson.bundledDependencies,
-      module: jsEntryPoint,
-      style: cssEntryPoint,
+      module: jsEntryPoint
+        ? path.join(paths.publicUrlOrPath, jsEntryPoint)
```
cangarugula, line 234: I recommend path.join but it actually doesn't resolve relative paths... I dug a little deeper and you don't need to do path.join it seems because `path.publicUrlOrPath` has a sanitized version of `PUBLIC_URL` so if you concat them together it should be fine. 

I still think you should keep ternary so that it's set to `undefined` if no jsEntrypoint is available.

cristiano-belloni: Done + added a test so we're sure the corner case doesn't generate double slashes

PR #1738, opened 2022-05-18
fix: fix modular ux and bugs

benpryke:


packages/modular-scripts/src/addPackage.ts
```diff
@@ -16,82 +16,138 @@ import LineFilterOutStream from './utils/LineFilterOutStream';
 import { ModularPackageJson } from './utils/isModularType';
 
 const packagesRoot = 'packages';
+const CUSTOM_TEMPLATE = '__CHOOSE_MY_OWN__';
 
 interface AddOptions {
-  destination: string;
-  type: string | void;
   name: string | void;
+  type: string | void;
```
cristiano-belloni, line 23: What's the benefit of using `string | void` instead of optionality (`?`) ?

benpryke: Good question. I think the difference is that the field must be present in the object, but can have an undefined value, rather than simply being missing from the object entirely.

If there's a benefit there, it's marginal. I copied the pattern that was already used rather than changing it though.

packages/modular-scripts/src/addPackage.ts
```diff
@@ -16,82 +16,138 @@ import LineFilterOutStream from './utils/LineFilterOutStream';
...
+  // so we loop until they choose a unique name or cancel
+  while (true) {
+    // Default the response the predefined `name` argument
+    const response = name
+      ? { name }
+      : ((await prompts({
+          name: 'name',
+          type: 'text',
+          message: `What would you like to name this package?`,
+        })) as { name: string } | Record<string, never>);
```
cristiano-belloni, line 41: Why the `Record<string, never>`?

cristiano-belloni (approved):
lgtm

packages/modular-scripts/src/addPackage.ts
benpryke: If the user cancels the prompt with `Ctrl+C`, we get an empty object back, literally just `{}`. I've used `Record<string, never>` to represent that empty object and allow us to check `if (!response.name)` on line 44, covering both the case where the user enters no name by pressing enter or presses `Ctrl+C`.

PR #1757, opened 2022-05-20
Support CSS from CDN in esm-views

cristiano-belloni:
## Problem

Suppose we have a CSS import from an external dependency like:

```ts
import 'regular-table/dist/css/material.css';
```

In an `esm-view`, we want to load the dependency from CDN instead of bundling it from local dependencies. In the future, we will also be able to use [CSS Module Scripts](https://web.dev/css-module-scripts/) to add styles to the page (and possibly deduplicate them).
The problem, with Webpack, is that we can't just rewrite the import like we do with js modules. We need to use an helper function that fetches the css (or [imports it with an assertion](https://web.dev/css-module-scripts/)) from the CDN.

## Solution

I used a [Webpack pitching loader](https://webpack.js.org/api/loaders/#pitching-loader) on top of our chain of style loaders. A pitching loader has a `pitch` method that gets evaluated left-to-right in the array of loaders. If it returns something truthy, it prevents the subsequent loaders to run; otherwise, all the other loaders for the rule run right-to-left (so the top loader is last). It works in our case because:

- If we detect that a CSS import is within our rewritable dependency map in the pitch method, we can rewrite it with some code that fetches and injects the desired file from CDN and, by returning from pitch, prevent all the other style loaders to process it.
- In all other cases (local CSS files or external CSS files that need to be bundled), we return `undefined` from the pitch function, letting the whole loader chain process it, and when our loader runs last, we just act as a bypass and return whatever output the loader chain has processed before us.

The pros of this approach is that it uses the correct mechanism to load a dependency (a loader) and the solution is quite compact. The cons are that we re-create the import submodule "manually" (by looking at the Webpack relative path of the requested file) and that the loader is mandatorily positional (it must sit before the others).

## Alternative solutions

- plugins: Plugins are more powerful than loaders. Theoretically we could scan the AST of every file and carefully rewrite it to fetch styles from a CDN where needed. In practice, it would be an incredibly complex solution that is much more difficult to debug.
- non-pitching loader: if it existed a way to "redirect" an import to a loader, using the [inline syntax](https://webpack.js.org/api/loaders/#inline-matchresource) and a query to specify the import statement value, we could use a much simple standalone loader. Unfortunately, it doesn't seem possible to do that when [marking imports as externals](https://webpack.js.org/configuration/externals/#function). Any help would be appreciated here.

## Other

This PR also includes a bit more debug logging when filtering external dependencies and a small bugfix (wildcard value in allow list is `**` instead of `*` to include scoped packages)


packages/modular-scripts/react-scripts/config/cdnStyleLoader.js
```diff
@@ -0,0 +1,37 @@
...
+      : dependency;
+    return generateStyleInjector(dependencyUrl);
+  }
+};
+
+function generateStyleInjector(url) {
+  return `
+  const link = document.createElement('link');
+  link.rel = 'stylesheet'; 
+  link.type = 'text/css';
```
cangarugula, line 33: What about `.less` files?

cristiano-belloni: I don't think we're expecting to process any files coming from the CDN at runtime - the expectation is that the CDN provides files that are already processed by the build process of the dependency.

packages/modular-scripts/react-scripts/config/cdnStyleLoader.js
```diff
@@ -0,0 +1,37 @@
...
+  link.href = '${url}'; 
+  document.getElementsByTagName('HEAD')[0].appendChild(link);
```
benpryke, line 35: Edge case: the `<head>` tag is optional, so it may not be present. I suggest creating it if it doesn't exist. https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -464,12 +477,17 @@ module.exports = function (webpackEnv) {
...
-              }),
+              use: getStyleLoaders(
+                {
+                  importLoaders: 1,
+                  sourceMap: isEnvProduction
+                    ? shouldUseSourceMap
+                    : isEnvDevelopment,
+                },
+                undefined,
+                true,
```
benpryke, line 488: Should this be `isEsmView`?

packages/modular-scripts/react-scripts/config/webpack.config.js
```diff
@@ -102,8 +102,18 @@ module.exports = function (webpackEnv) {
...
 
   // common function to get style loaders
-  const getStyleLoaders = (cssOptions, preProcessor) => {
+  const getStyleLoaders = (cssOptions, preProcessor, includeEsmLoader) => {
     const loaders = [
+      // This loader translates external css dependencies if we're using a CDN
+      // Since it's a pitching loader, it's important that it stays at the top
+      // excluding all the others in the chain if it's triggered
+      includeEsmLoader &&
+        function externalStyleLoader(info) {
```
benpryke, line 111: There's already an import for `info` on line 17 of this file. I know this new variable is scoped to this function, but I wonder whether its name would be clearer if it were more specific.

packages/modular-scripts/src/__tests__/esmView.test.ts
```diff
@@ -195,8 +195,7 @@ describe('modular-scripts', () => {
         stdio: 'inherit',
         env: {
           USE_MODULAR_ESBUILD: 'true',
-          EXTERNAL_CDN_TEMPLATE:
-            'https://mycustomcdn.net/[name]?version=[version]',
+          EXTERNAL_CDN_TEMPLATE: 'https://mycustomcdn.net/[name]@[version]',
```
benpryke, line 198: nit: you've used this string about 5 times, maybe just use a variable for `EXTERNAL_CDN_TEMPLATE`

packages/modular-scripts/src/__tests__/esmView.test.ts
```diff
@@ -310,19 +305,97 @@ describe('modular-scripts', () => {
...
+      expect(
+        prettier.format(indexFile, {
+          filepath: 'index-F6YQ237K.js',
+        }),
+      ).toMatchSnapshot();
+      expect(indexFile).toContain(`https://mycustomcdn.net/react@1`);
+      expect(indexFile).toContain(`https://mycustomcdn.net/lodash@^4.17.21`);
       expect(indexFile).toContain(
-        `https://mycustomcdn.net/lodash.merge?version=^4.6.2`,
+        `https://mycustomcdn.net/lodash.merge@^4.6.2`,
```
benpryke, line 398: Should we be using hats or exact versions for consistency at runtime?

packages/modular-scripts/react-scripts/config/cdnStyleLoader.js
```diff
@@ -0,0 +1,37 @@
...
+// For this reason, this loader must be first in the chain of other style loaders.
+// It will just set the output and interrupt if pitch returns and it will just act as a bypass of the chain output otherwise
+// If it existed another method of redirecting an external to a loader, it'd be simpler to just get rid of these pitch restrictions
+
+module.exports = function cdnStyleLoader(asset) {
+  // noop. Just return the translated asset if the pitching doesn't return (and the chain is processed)
+  return asset;
+};
+
+module.exports.pitch = function () {
```
benpryke, line 13: Being unfamiliar with pitching loaders and the details here, I do not claim to fully understand the implications of this implementation. Do you think the tests you have provided cover the edge cases that might occur?

cristiano-belloni: We decided to add a test for scoped packages in allow / block list. (test the behaviour of `filterDependencies` in `utils`)

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: We decided to not destructure in line 17 (`const { info } = require('../../react-dev-utils/logger');`)

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: Done, thanks!

packages/modular-scripts/src/__tests__/esmView.test.ts
cristiano-belloni: Hats, we already have a test for exact versions (resolutions)

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: Done

packages/modular-scripts/react-scripts/config/cdnStyleLoader.js
cristiano-belloni: Done.

packages/modular-scripts/react-scripts/config/webpack.config.js
cristiano-belloni: Done

packages/modular-scripts/src/__tests__/esmView.test.ts
cristiano-belloni: We said that we wanted to leave it like this because it's more immediate to read and the tests (we have to be DRY but tests need to be legible more)

packages/modular-scripts/src/__tests__/esmView.test.ts
```diff
@@ -737,7 +799,83 @@ describe('modular-scripts', () => {
             path.join(modularRoot, 'dist', 'sample-esm-view', 'package.json'),
           )) as CoreProperties
         ).module,
-      ).toEqual('./static/js/index-5PFA727O.js');
+      ).toEqual('./static/js/index-CWZPL7BC.js');
+    });
+  });
+
+  describe('WHEN we specify an allow / block list', () => {
```
benpryke, line 806: This test file is very long, and these unit tests may get lost. I think it would be better to put them in a file that makes it clear this is testing `filterDependencies.ts`.

packages/modular-scripts/src/utils/filterDependencies.ts
```diff
@@ -18,22 +19,54 @@ export function filterDependencies(
...
+    packageDependencies,
+    allowList: externalAllowList,
+    blockList: externalBlockList,
+  });
+}
+
+export function matchDependencies({
+  packageDependencies,
+  // By default, everything in allow list
+  allowList = ['**'],
```
benpryke, line 54: Re-reading this, I think "allow", "block", and even "matchDependencies" (and "filterDependencies") are not self-descriptive. Really what you're looking for is external vs bundled. It'll be very easy to forget what this function is trying to do or what it's for.

cristiano-belloni: @benpryke I changed terminology to reflect extarnal / bundled in the function that was introduced in this PR, and renamed it to `partitionDependencies`.
I didn't change it in the rest of the file, since `EXTERNAL_BLOCK_LIST` and `EXTERNAL_ALLOW_LIST` were approved months ago, are part of the user interface and not in scope of this PR.

packages/modular-scripts/src/__tests__/esmView.test.ts
benpryke: That's fair enough, though I would have been happy if you'd created a to-be-completed test file as we can always flesh it out later!

benpryke (approved):

PR #1766, opened 2022-05-25
fix: ensure transpiled ts is compatible with babel

benpryke:
Users are still running into the issue where the output from the TypeScript compiler is not compatible with our babel/webpack config. This causes errors at build time for new language features such as the optional chaining operator.

Although it does fix the issue, using "last 30 chrome versions" (etc) is a bit arbitrary so I'm open to other suggestions!

This would be a patch release.


cristiano-belloni:
> Users are still running into the issue where the output from the TypeScript compiler is not compatible with our babel/webpack config. This causes errors at build time for new language features such as the optional chaining operator.

What is the repro? Using optional chaining would break the build?

> This would be a patch release.

Could you please add a patch changeset then?

LukeSheard:
I'd suggest that there's a test suite for these various language features to make sure that out the box `modular build` works with them as either an app of view.

cristiano-belloni (dismissed):
👍 This needs to be expanded into a test suit, but approving it in the meanwhile.

cristiano-belloni (dismissed):

PR #1772, opened 2022-05-26
Chore/replace yarnpkg with yarn

cangarugula:
yarnpkg` points to `yarn` by default and `yarnpkg` itself is deprecated so we should point to yarn directly instead of pointing to `yarnpkg`


cangarugula:
Closing this PR due to to issue https://github.com/facebook/create-react-app/pull/1365 

TLDR; Hadoop installs a bin named `yarn` which causes conflicts when executing `yarn` so we need to point to `yarnpkg` which is aliased to the `yarn` bin we want.

PR #1792, opened 2022-06-01
fix: support object type for package.json workspaces field

benpryke:
Although it is typical for the `workspaces` field in `package.json` to be an array, it is also valid for it to be an object as described in yarn's blog: https://classic.yarnpkg.com/blog/2018/02/15/nohoist/. If a modular user adds the `nohoist` field to their root `package.json` file, `modular check` will crash.

This PR fixes that and ensures that all the code that validates the `package.json` structure that modular expects lives in the same place.


packages/modular-scripts/src/utils/isModularType.ts
```diff
@@ -19,14 +21,31 @@ export type PackageType = ModularTemplateType | 'template';
 
 export type ModularType = PackageType | 'root';
 
-export type ModularPackageJson = PackageJson & {
+// Utility type that extends type `T1` with the fields of type `T2`
+type Extend<T1, T2> = {
```
steveukx, line 25: 🎉

steveukx (dismissed):

packages/modular-scripts/src/check/verifyModularRootPackageJson.ts
```diff
@@ -0,0 +1,62 @@
...
+import getModularRoot from '../utils/getModularRoot';
+import { ModularPackageJson } from '../utils/isModularType';
+import * as logger from '../utils/logger';
+
+/**
+ * @param rootPackageJson package.json object from the modular root
+ * @returns Returns true if the packages directory is correctly included in `workspaces`, else false
+ */
+function checkWorkspaceHasPackages(rootPackageJson: ModularPackageJson) {
+  if (rootPackageJson?.workspaces) {
```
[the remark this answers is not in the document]
benpryke: In practice, it depends on how the data passed in is read and typed. If we were to pass in the result of

    fs.readJsonSync(packageJsonPath, { throws: false }) as ModularPackageJson

there would be no type errors, but `rootPackageJson` could be `null`.

If I type the parameter with `ModularPackageJson | undefined`, I'm signalling that `undefined` is a valid input, but it is in fact undesirable. In fact, our current inputs will never be null, yet I cannot guard against future changes, so I coded defensively.

In general, I dislike how TypeScript encourages defenceless code; are there better practices I should follow?

packages/modular-scripts/src/check/verifyModularRootPackageJson.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ModularPackageJson } from '../utils/isModularType';
+import * as logger from '../utils/logger';
+
+/**
+ * @param rootPackageJson package.json object from the modular root
+ * @returns Returns true if the packages directory is correctly included in `workspaces`, else false
+ */
+function checkWorkspaceHasPackages(rootPackageJson: ModularPackageJson) {
+  if (Array.isArray(rootPackageJson?.workspaces)) {
+    return !!rootPackageJson.workspaces.includes('packages/**');
```
cangarugula, line 15: `.includes` returns a boolean so you don't need the double bang

packages/modular-scripts/src/check/verifyModularRootPackageJson.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  return !!rootPackageJson?.workspaces?.packages?.includes('packages/**');
```
cangarugula, line 18: Same here

cristiano-belloni: This one can evaluate to `undefined` though.

cristiano-belloni (approved):

PR #1793, opened 2022-06-01
Fix: Correctly parse v3 lockfiles

cristiano-belloni:
## Problem

- `@yarnpkg/lockfile` is able to parse lockfile v1 format correctly but not lockfile v3 format (it throws immediately)
- The tools based on `@yarnpkg/lockfile` that we were using error with the lockfile v3 format
- It seems that lockfiles v2+ are not supported because they are valid yaml

## Solution

- Try to parse lockfile directly with `@yarnpkg/lockfile`
- If it crashes, parse it manually from YAML.


packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -69,26 +71,41 @@ export async function getPackageDependencies(
...
+      {},
+    );
+  } catch (e) {
+    // Try to parse as yaml (v2+) - https://github.com/yarnpkg/yarn/issues/5629#issuecomment-753418765
+    const parsedLockfile = yaml.load(lockFile) as LockFileEntries;
+    const dependencyArray = Object.entries(deps);
+    // Parse: yarn v3 lock comes with entries like "'yargs@npm:^15.0.2, yargs@npm:^15.1.0, yargs@npm:^15.3.1, yargs@npm:^15.4.1'"
+    Object.entries(parsedLockfile).forEach(([name, { version }]) => {
+      const entryDependencies = name.split(', ');
+      dependencyArray.some(([dependencyName, dependencyVersion]) => {
```
benpryke, line 95: As you're not actually using the return value of `Array.prototype.some`, a for loop with a break statement would be clearer.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -69,26 +71,41 @@ export async function getPackageDependencies(
...
+          `${name}@${version}`
+        ].version;
+        return acc;
+      },
+      {},
+    );
+  } catch (e) {
+    // Try to parse as yaml (v2+) - https://github.com/yarnpkg/yarn/issues/5629#issuecomment-753418765
+    const parsedLockfile = yaml.load(lockFile) as LockFileEntries;
+    const dependencyArray = Object.entries(deps);
```
benpryke, line 91: What's the difference between `dependencyArray` and `Object.entries(parsedLockfile)`? I don't quite understand what this `O(n^2)` algo is doing.

benpryke: Why do we not need to add the dependencies to the right of the `parsedLockfile` entries after a match is found?

When you extract this to a separate function, I'd recommend adding a little context so it's easier for the next person who reads this to get to grips with the why. I'm not familiar with the new lockfile format!

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -69,26 +71,41 @@ export async function getPackageDependencies(
...
-        ) as Dependency,
-      }),
-    ),
-    lockFile,
-    true,
-    LockfileType.yarn,
-  );
+  let lockDeps: Dependency = {};
+
+  try {
```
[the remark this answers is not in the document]
cristiano-belloni: Done

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -69,26 +71,41 @@ export async function getPackageDependencies(
...
+            `${dependencyName}@npm:${dependencyVersion}`,
+          )
+        ) {
+          lockDeps[dependencyName] = version;
+          return true;
+        }
+        return false;
+      });
+    });
+    // Prepare the parsed lockfile for access
```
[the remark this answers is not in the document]
cristiano-belloni: Done

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: I find it quite idiomatic when we ignore the return value (functional breaking loop), but no prob, done.

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Refactored + added additional comments.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -120,3 +103,46 @@ export async function getPackageDependencies(
     );
   return { manifest, resolutions };
 }
+
+function parseYarnLock(lockFile: string, deps: Dependency): Dependency {
+  return parseYarnLockV1(lockFile, deps) || parseYarnLockV3(lockFile, deps);
```
cangarugula, line 108: Since we are explicitly catering to yarn 1 and yarn 3, can we add this to `isYarnInstalled` in `startupCheck.ts` ?

cangarugula: By this, I mean the check for the version of yarn to be explicit with our compatibility

cangarugula: Spoke offline and decided to create a separate ticket to resolve yarn version compatibility

cangarugula (approved):

PR #1821, opened 2022-06-08
Fix package resolutions with workspace dependencies

cristiano-belloni:
- `modular analyze` will show workspace dependencies with their correct resolutions (before it didn't)
- `modular build` and `modular start` will build (and bundle / not rewrite by default) the workspace dependencies the target package depends on (before those commands were breaking because `getDependenciesInfo` didn't find the resolutions in `yarn.lock`)
- Debug logging is more thorough and comprehensible
- Conditional yarnv1 / yarnv3 lockfile parsing error reporting works better (before, _any_ exception in the v1 parsing block would trigger a v3 parse, with consequent confusing "parse error" message - now it reports the correct error)
- Support `workspace:` deps
- Tests added
- Docs updated with mention of how we bundle workspace dependencies by default

### DIFF
--- a/packages/modular-scripts/src/__tests__/utils/getPackageDependencies.test.ts
+++ b/packages/modular-scripts/src/__tests__/utils/getPackageDependencies.test.ts
@@ -1,0 +1,164 @@
+import {
+  parseYarnLock,
+  resolvePackageDependencies,
+} from '../../utils/getPackageDependencies';
+
+describe('Resolve dependencies', () => {
+  it('should resolve dependencies', () => {
+    const { manifest, resolutions, manifestMiss, lockFileMiss } =
+      resolvePackageDependencies({
+        dependenciesfromSource: ['dependency-1', 'dependency-2'],
+        packageDeps: {
+          'dependency-1': '^1.1.2',
+          'dependency-2': '^2.3.4',
+          'dependency-3': '>4.5.6',
+        },
+        lockDeps: {
+          'dependency-1': '1.1.8',
+          'dependency-2': '2.4.0',
+          'dependency-3': '4.9.8',
+        },
+        workspaceInfo: {
+          'other-dependency': {
+            location: '',
+            version: '3.0.0',
+            workspaceDependencies: [],
+            mismatchedWorkspaceDependencies: [],
+            type: 'esm-view',
+            public: false,
+          },
+        },
+      });
+    expect(manifest).toEqual({
+      'dependency-1': '^1.1.2',
+      'dependency-2': '^2.3.4',
+    });
+    expect(resolutions).toEqual({
+      'dependency-1': '1.1.8',
+      'dependency-2': '2.4.0',
+    });
+    expect(lockFileMiss).toEqual([]);
+    expect(manifestMiss).toEqual([]);
+  });
+
+  it('should resolve dependencies with pacxkage and lockfile misses', () => {
+    const { manifest, resolutions, manifestMiss, lockFileMiss } =
+      resolvePackageDependencies({
+        dependenciesfromSource: [
+          'dependency-1',
+          'dependency-2',
+          'dependency-missing',
+        ],
+        packageDeps: {
+          'dependency-1': '^1.1.2',
+          'dependency-2': '^2.3.4',
+          'dependency-3': '>4.5.6',
+        },
+        lockDeps: {
+          'dependency-1': '1.1.8',
+          'dependency-2': '2.4.0',
+          'dependency-3': '4.9.8',
+        },
+        workspaceInfo: {
+          'other-dependency': {
+            location: '',
+            version: '3.0.0',
+            workspaceDependencies: [],
+            mismatchedWorkspaceDependencies: [],
+            type: 'esm-view',
+            public: false,
+          },
+        },
+      });
+    expect(manifest).toEqual({
+      'dependency-1': '^1.1.2',
+      'dependency-2': '^2.3.4',
+    });
+    expect(resolutions).toEqual({
+      'dependency-1': '1.1.8',
+      'dependency-2': '2.4.0',
+    });
+    expect(lockFileMiss).toEqual(['dependency-missing']);
+    expect(manifestMiss).toEqual(['dependency-missing']);
+  });
+});
+
+describe('Parse yarn lockfiles', () => {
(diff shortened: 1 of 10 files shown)


packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -112,20 +114,24 @@ function parseYarnLockV1(
...
-    return Object.entries(deps).reduce<Record<string, string>>(
-      (acc, [name, version]) => {
-        acc[name] = (parsedLockfile.object as LockFileEntries)[
-          `${name}@${version}`
-        ].version;
-        return acc;
-      },
-      {},
-    );
+    parsedLockfile = lockfile.parse(lockFile);
```
steveukx, line 119: This is a bit confusingly named - `lockile` and `lockFile`

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -112,20 +114,24 @@ function parseYarnLockV1(
...
+      if (resolution !== undefined) {
+        acc[name] = resolution;
+      }
+      return acc;
+    },
+    {},
+  );
 }
 
 function parseYarnLockV3(lockFile: string, deps: Dependency): Dependency {
```
steveukx, line 137: Not in your PR - but this function may be more readily understood by limiting the loops inside loops by preparing a look-up Map before looping through the lockfile entries:

```typescript
function parseYarnLockV3(lockFile: string, deps: Dependency): Dependency {
  const dependencyArray = Object.entries(deps);
  const npmDependencyList = new Map(
    dependencyArray.map(([name, version]) => [`${name}@npm:${version}`, name])
  );
  const resolved: Dependency = {};

  Object.entries(yaml.load(lockFile) as LockFileEntries).forEach(
    ([name, { version }]) => {
    // Yarn v3 lockfile comes with keys like "'yargs@npm:^15.0.2, yargs@npm:^15.1.0, yargs@npm:^15.3.1, yargs@npm:^15.4.1'" - split them

    const entryDependencies = name.split(', ');
    entryDependencies.some(dep => {
      const npmName = npmDependencyList.get(dep);
      return !!(npmName && (resolved[npmName] = version));
    });
  });

  return resolved;
}
```

Exporting this function from a utility file would then permit unit testing directly without the need to rely on a surrounding integration test.

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Yep - renamed it with a more descriptive / less confusing name.

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: I changed `some` to a `for` loop based on an indication [in a previous PR](https://github.com/jpmorganchase/modular/pull/1793#discussion_r887037489). TBH, I like the pre-calculation of `npmDependencyList` and I find using `reduce` and `some` more compact than explicitly looping and breaking.

+1 for unit testing those functions.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -109,30 +114,37 @@ function parseYarnLock(lockFile: string, deps: Dependency): Dependency {
...
+      const resolution = (parsedLockfile.object as LockFileEntries)[
+        `${name}@${version}`
+      ]?.version;
+      if (resolution !== undefined) {
+        acc[name] = resolution;
+      }
+      return acc;
+    },
+    {},
+  );
```
steveukx, line 137: As the inline type assertion splits this over several lines, you could flatten this by making the assertion outside of the `.reduce`:

```typescript
const lockFileEntries = parsedLockfile.object as LockFileEntries;
  return Object.entries(deps).reduce<Record<string, string>>(
    (acc, [name, version]) => {
      const key = `${name}@${version}`;
      if (lockFileEntries[key]) {
        acc[name] = lockFileEntries[key].version;
      }
      return acc;
    },
    {},
  );
```

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Thanks, done (but maintaining the reducer)

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: 👍 done

packages/modular-scripts/src/__tests__/utils/getPackageDependencies.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    });
+    expect(resolutions).toEqual({
+      'dependency-1': '1.1.8',
+      'dependency-2': '2.4.0',
+    });
+    expect(lockFileMiss).toEqual([]);
+    expect(manifestMiss).toEqual([]);
+  });
+
+  it('should resolve dependencies with pacxkage and lockfile misses', () => {
```
steveukx, line 44:
```suggestion
  it('should resolve dependencies with package and lockfile misses', () => {
```

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -58,91 +66,137 @@ export async function getPackageDependencies(
...
-        }
-        return acc;
-      },
-      { manifest: {}, resolutions: {} },
-    );
-  return { manifest, resolutions };
+  // Log the errors
+  resolvedPackageDependencies.manifestMiss.forEach((depName) =>
+    logger.error(
+      `Package ${depName} imported in ${target} source but not found in package dependencies or hoisted dependencies - this will prevent you from successfully build, start or move esm-views and will cause an error in the next release of modular`,
```
sgb-io, line 98: nit: grammar x2 - "from successfully building, starting or moving esm-views"?

sgb-io (dismissed):

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Good catch, done!

sgb-io (approved):

PR #1867, opened 2022-06-23
Splitting up large processes

steveukx:
Split the main `modular build` handler into a lifecycle style runner,
 each step of the process can be handled by one or more async functions
 run in series, sharing build state with a `StandAloneBuilderContext`.

Configuring the steps and running the build is now split into separate
 concerns, which should help run builds concurrently as their state
 isn't shared.


packages/modular-scripts/src/build/fileSizeReporter.ts
```diff
@@ -3,6 +3,7 @@ import chalk from 'chalk';
 import filesize from 'filesize';
 import stripAnsi from 'strip-ansi';
 import * as logger from '../utils/logger';
+import { StandAloneBuilderContext } from './createBuilderContext';
```
cristiano-belloni, line 6: `import type`?

PR #1917, opened 2022-07-08
Refactor esm-view tests to be platform-independent

cristiano-belloni:
This PR refactors esm-view tests to work on Windows (and adds a Github action for the Windows platform). Snapshot testing is preserved and improved, but tests are much more resilient to subtle differences (such as line separators or path separators)

**_Remember to expand `packages/modular-scripts/src/__tests__/esmView.test.ts` in the diff view, it's too long to be automatically shown by Github_**

- **Remove inline snapshot of the generated directory tree, with file hashes**, because:
  - These tests don't provide much value: They say something is changed, but they don't say how (they don't show a diff)
  - They prevent multi platform testing: since some filenames are hashed, there are cases in which filenames will be inevitably generated differently on different platforms (for example: sourcemaps line separators are platform dependent and they generate different hashes)
- **Strip source map comment statements generated js files**. There is no way for source maps to be hashed uniformly, since their embedded strings will have different line separators on different OSes. We still need to snapshot generated files, but after removing any reference to sourcemaps.
- **Snapshot generated js files explicitly**. This is useful because, when a test fails, we can see why directly from the log (diff is printed). Before, we had to checkout the previous version, save the generated file, then diff it with latest.
- **Remove hardcoded file names**. Before, whenever a the has part in a generated file name failed, we had to manually replace all its occurrences through the test file. Now, we just need to check what's changed in the diff (se previous point) and, if the change is expected, run `yarn test -u` to update the snapshots.
- **Kill develop server in a portable way**. Stop using `lsof`, which is not present on Windows. Use `kill-port` from npm in its place.


sgb-io:
Not sure if it's expected, but I get a test error on my win11 / node16 device:

```
 FAIL   test  packages/sample-esm-view/src/__tests__/EsmView.test.tsx
  ● Test suite failed to run

    ENOENT: no such file or directory, open 'C:\Users\sam\Repos\modular\packages\sample-esm-view\src\__tests__\EsmView.test.tsx'

      at runTestInternal (../node_modules/jest-runner/build/runTest.js:202:27)

Test Suites: 1 failed, 2 passed, 3 total
Tests:       31 passed, 31 total
Snapshots:   14 passed, 14 total
Time:        105.647 s
Ran all test suites matching /esmView.test.ts/i.
[modular] ✕ Modular test did not pass
error Command failed with exit code 1.
```

cristiano-belloni:
> Not sure if it's expected, but I get a test error on my win11 / node16 device:

This usually happens when the `sample-esm-view` package doesn't get removed from the repo, either because you interrupted the test manually or because the tests are not properly cleaning it up. I'll try on my Windows machine too.

sgb-io:
> > Not sure if it's expected, but I get a test error on my win11 / node16 device:
> 
> This usually happens when the `sample-esm-view` package doesn't get removed from the repo, either because you interrupted the test manually or because the tests are not properly cleaning it up. I'll try on my Windows machine too.

Yeah, agreed, it's not a real issue

sgb-io (approved):
LGTM

PR #1978, opened 2022-07-27
Implement @modular-scripts/workspace-resolver

sgb-io:
Implements a manual replacement for the yarn v1 version of `yarn workspaces info` that will work across versions of yarn.

The output of `analyzeWorkspaceDependencies` is compatible with the yarn v1 version of `yarn workspaces info` with the exception that yarn >= v2 workspace ranges are also supported (whereas the yarn v1 `yarn workspaces info` command would see all workspace ranges as mismatched versions).

## Summary

- Implements 2 new packages: `@modular-scripts/workspace-resolver` and `@modular-scripts/modular-types`
- `@modular-scripts/workspace-resolver` exposes `resolveWorkspace` and `analyzeWorkspaceDependencies` (documentation in `packages/workspace-resolver/README.md`)
- Updated GitHub CI steps as needed

## Decisions taken

- Enabled the updateInternalDependents option to support publishing inter-dependencies as we want
  - See: https://github.com/changesets/changesets/blob/main/docs/experimental-options.md#updateinternaldependents-type-out-of-range--always
  - See: https://github.com/changesets/changesets/issues/476 and https://github.com/changesets/changesets/pull/542
- Added types: In order to implement `@modular-scripts/workspace-resolver` as a publishable package, it was necessary to implement `@modular-scripts/modular-types`. Without this, a circular dependency happens.
- Added some test fixtures to the root of the project: these were originally inside the `@modular-scripts/workspace-resolver` package, but this causes them to be detected by yarn and other tools when globbing for packages in `packages/**`. Changing the `workspaces` glob was unsuccessful, as we have tests that rely on the `packages/**` value (including workspaces that get created inside `__tests__` directories, within `packages`). So, the cleanest way to store these tests is outside of `packages`. I also decided it's best to store these fixtures statically in the repo instead of generating them on the fly in tests


.github/workflows/test-windows.yml
```diff
@@ -31,5 +31,9 @@ jobs:
           cache: 'yarn'
       - name: Install Dependencies
         run: yarn --frozen-lockfile
-      - name: Run Windows tests
+      - name: Build prerequisites
+        run: yarn workspace modular-types build
+      - name: Run Windows tests (part 1)
         run: yarn test esmView.test.ts
+      - name: Run Windows tests (part 2)
+        run: yarn test workspace-resolver
```
cristiano-belloni, line 39: Can this be combined in the same step? I think you can simply do

`yarn test esmView.test.ts workspace-resolver`

packages/modular-types/src/types.ts
```diff
@@ -0,0 +1,26 @@
+export type ModularTemplateType = 'app' | 'esm-view' | 'view' | 'package';
```
cristiano-belloni, line 1: If it's only types there, can't you just write a .d.ts file yourself, point to it in the package.json and don't have to build it? It will be not much different than what you have here, but with much less tooling complexity (yes, I'm advocating for import from source, but in this case there's no real build step)

sgb-io: Good point, I hadn't thought of it this way, but I guess there is no reason *not* to, if the only thing that gets published under this package is types.

That said, is that the same as importing from source? I think we can still have this as a distinct package rather than a direct folder of types

cristiano-belloni: > That said, is that the same as importing from source? I think we can still have this as a distinct package rather than a direct folder of types

It should still be a workspace with its own package.json and the "types" field, only sans build step.

__fixtures__/invalid-workspace-1/packages/package-two/src/index.ts
```diff
@@ -0,0 +1 @@
+console.log('Package two');
```
cristiano-belloni, line 1: Do we really need these source files? Technically we don't even look at them right?

packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -0,0 +1,187 @@
...
+      .filter(([dep, range]) => {
+        const matchingPackage = packageDeps.find(
+          ([matchingPackageName]) => dep === matchingPackageName,
+        );
+        if (!matchingPackage) {
+          return false;
+        }
+
+        const [, match] = matchingPackage;
+        return !semver.satisfies(match.version, range);
```
cristiano-belloni, line 173: Could we support [workspace ranges](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace) too here? They're heavily used in Yarn >= v2 and they might be "standardised" in the future (ie supported by a npm RFC)

packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -0,0 +1,187 @@
...
+          cwd,
+        },
+      );
+    });
+  }
+
+  return resolveWorkspacesDefinition(cwd, def.packages);
+}
+
+type PackageJson = {
```
cristiano-belloni, line 40: We normally use [these definitions](https://www.npmjs.com/package/@schemastore/package) through the whole project. If you need something that's not in there, could you extend one of their types?

packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -0,0 +1,187 @@
...
+  dependencies: Record<string, string> | undefined;
+};
+
+function readPackageJson(
+  isRoot: boolean,
+  workingDir: string,
+  relativePath: string,
+): Promise<PackageJson> {
+  if (isRoot) {
+    return readJson(relativePath) as Promise<PackageJson>;
```
cristiano-belloni, line 56: nit: instead of rewriting the whole `readJson` signature twice, can we just conditionally create the filename that we're reading?

steveukx: 👍 shame `readJson` doesn't allow for generics, or supplying a schema to validate the JSON while it's being read...

packages/workspace-resolver/src/resolve-workspace.ts
sgb-io: Yep adding this, but: it should be noted, this makes `@modular-scripts/workspace-resolver` no longer 100% compatible with `yarn workspaces info` (which makes sense, because workspace ranges were not a thing in yarn v1).

cristiano-belloni (dismissed):

cristiano-belloni (approved):

cristiano-belloni (changes requested):
Remove modular-scripts from changeset

cristiano-belloni (approved):

cristiano-belloni (changes requested):
Let's pause this until we have a strategy for publishing.

cristiano-belloni (approved):

cristiano-belloni (approved):

cristiano-belloni (approved):

PR #1984, opened 2022-07-31
Support the new JSX transform

joshwooding:
- Bump esbuild from 0.14.42 to 0.14.51
- Switch to using "jsx": "react-jsx" essentially reverting https://github.com/jpmorganchase/modular/pull/502


joshwooding:
Test is failing because skypack isn't compatible with the new transform esm.sh is though. If we can be a bit less opinionated we can just rely on the jsx option in tsconfig which would fix the test but users might run into this in the future.

joshwooding:
Swapped to using a opt-in approach via the tsConfig to prevent errors when using the esm view

sgb-io:
Hi @joshwooding, a fixed version is open here: #1998

sgb-io:
Superseded by #1998

PR #2001, opened 2022-08-04
Integrate workspace resolver into modular-scripts

sgb-io:
### Backwards-compatible implementation swap for `getAllWorkspaces()`

- Implemented `@modular-scripts/workspace-resolver` in `getAllWorkspaces`. Also removed the old implementation.
- Updated the interface of `getAllWorkspaces` to accept an optional `target` string, which can be thought of as a working directory. It is required when the modular root based off `process.cwd()` is not the same as the directory we want to parse workspaces from. This allows the new implementation to work with operations such as `convert` and `port` without any breakages.
- Updated the interface of `Check` so that `check()` has an optional `target?: string` argument
- Changed `getWorkspaceInfo` to work with the new implementation *and* perform less filesystem IO
  - Since we are already reading `package.json` files in the workspace-resolver, reading them again would be duplicate filesystem IO work
  - By adding the `location` and `rawPackageJson` fields to `ModularWorkspacePackage` in `@modular-scripts/workspace-resolver`, we save all this extra IO
  - The approach with `rawPackageJson` can be considered "lazy": the alternatives as I see it are to either (a) keep the duplicate IO or (b) reduce it to _what we need_ - but this would require significant refactoring, as the whole content is currently exposed by `getPackageMetadata` & various commands depend on that function
- Changed `getPackageMetadata` to work with the new structures returned by `@modular-scripts/workspace-resolver`
- Also made `getPackageMetadata` use the more accurate `@modular-scripts/workspace-resolver` `ModularPackageJson` type instead of `JSONSchemaForNPMPackageJsonFiles`
- Renamed `getWorkspaceInfo` in `utils/getAllWorkspaces` to become `getWorkspacePackages`. This is because `getWorkspaceInfo` was conflicting with a function of the same name in `utils/getWorkspaceInfo`. It was quite confusing before.
- Updated CI so that `@modular-scripts/workspace-resolver` is built as a prerequisite step - this is needed because we're now using it directly in `modular-scripts`

### DIFF
--- a/packages/modular-scripts/src/__tests__/init.test.tsx
+++ b/packages/modular-scripts/src/__tests__/init.test.tsx
@@ -3,7 +3,7 @@ import * as path from 'path';
 import * as tmp from 'tmp';
 import * as fs from 'fs-extra';
 import { promisify } from 'util';
-import { getWorkspaceInfo } from '../utils/getAllWorkspaces';
+import { getWorkspacePackages } from '../utils/getAllWorkspaces';
 
 import type { ModularPackageJson } from '@modular-scripts/modular-types';
 
@@ -47,7 +47,7 @@ describe('Creating a new modular folder', () => {
   });
 
   it('should not have any workspace info', async () => {
-    const workspace = await getWorkspaceInfo(folder);
+    const workspace = await getWorkspacePackages(folder);
     expect(workspace).toEqual({});
   });
 });
--- a/packages/modular-scripts/src/__tests__/init.test.tsx
+++ b/packages/modular-scripts/src/__tests__/init.test.tsx
@@ -51,9 +51,12 @@ describe('Creating a new modular folder', () => {
 
     // 1 item in the workspace output (the root package)
     expect(workspace[0].size).toEqual(1);
-    const item = Array.from(workspace[0])[0][1];
-    expect(item.version).toEqual('1.0.0');
-    expect(item.workspace).toEqual(true);
+    const workspaceMap = workspace[0];
+    expect(workspaceMap).toEqual(expect.any(Map));
+    const [, workspaceContent] = Array.from(workspaceMap)[0];
+    expect(workspaceContent).toEqual(
+      expect.objectContaining({ version: '1.0.0', workspace: true }),
+    );
 
     // Empty dependency analysis
     expect(workspace[1]).toEqual({});
--- a/packages/modular-scripts/src/__tests__/init.test.tsx
+++ b/packages/modular-scripts/src/__tests__/init.test.tsx
@@ -46,8 +46,16 @@ describe('Creating a new modular folder', () => {
     expect(String(lockfile)).toMatchSnapshot();
   });
 
-  it('should not have any workspace info', async () => {
-    const workspace = await getWorkspacePackages(folder);
-    expect(workspace).toEqual({});
+  it('should contain 1 workspace (the root) and an empty dependency analysis', async () => {
+    const workspace = await getWorkspacePackages(folder, folder);
+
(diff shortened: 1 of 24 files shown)


packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -62,13 +62,14 @@ export async function resolveWorkspace(
 > {
   const workingDirToUse = workingDir ?? process.cwd();
   const isRoot = workingDirToUse === root;
-  const path = packageJsonPath(root);
-  const json = await readPackageJson(isRoot, workingDirToUse, path);
+  const pkgPath = packageJsonPath(root);
```
sgb-io, line 65: Renamed `path` => `pkgPath` because `path` conflicts with the internal node module

packages/modular-scripts/src/convert.ts
```diff
@@ -164,7 +164,7 @@ export async function convert(cwd: string = process.cwd()): Promise<void> {
     });
 
     logger.log('Validating your modular project...');
-    await check();
+    await check(false, cwd);
```
cristiano-belloni, line 167: This is a bit difficult to read. Can we change all interfaces to
`{ fix: boolean, targetDirectory?: string }`

packages/modular-scripts/src/__tests__/init.test.tsx
```diff
@@ -46,8 +46,16 @@ describe('Creating a new modular folder', () => {
...
-  it('should not have any workspace info', async () => {
-    const workspace = await getWorkspaceInfo(folder, 'yarn');
-    expect(workspace).toEqual({});
+  it('should contain 1 workspace (the root) and an empty dependency analysis', async () => {
+    const workspace = await getWorkspacePackages(folder, folder);
+
+    // 1 item in the workspace output (the root package)
+    expect(workspace[0].size).toEqual(1);
+    const item = Array.from(workspace[0])[0][1];
+    expect(item.version).toEqual('1.0.0');
```
steveukx, line 55: These might be more expressive with the `expect.objectContaining`:

```
expect(Array.from(workspace[0])).toEqual(
  [
    expect.any(Map),
    expect.objectContaining({ version: '1.0.0', workspace: true })
  ]
)
```

packages/modular-scripts/src/check/index.ts
```diff
@@ -10,24 +10,29 @@ const CHECKS = [
...
+  target?: string;
 }
 
-export async function check(fix = false): Promise<void> {
+interface Verification {
+  check(target?: string): Promise<boolean>;
+  fix?(): Promise<void>;
+}
+
+export async function check({ fix = false, target }: Check): Promise<void> {
```
steveukx, line 22: For a consistent approach to declaring defaults, the default for `target` can also be defined in the function signature:

```suggestion
export async function check({ fix = false, target = process.cwd() }: Check): Promise<void> {
```

packages/modular-scripts/src/utils/getAllWorkspaces.ts
```diff
@@ -1,113 +1,35 @@
...
-    if (yarnVersion.startsWith('3.')) {
-      return supportedPackageManagers.yarn3;
-    }
-  }
-
-  throw new Error(`${packageManager} is not supported.`);
-}
+export async function getWorkspacePackages(
+  modularRoot: string,
+  target?: string,
```
steveukx, line 22: If `target` isn't supplied but the `modularRoot` is, should the default value of `target` also be `modularRoot`?

```suggestion
  target?: string = modularRoot,
```

cristiano-belloni (approved):

PR #2015, opened 2022-08-09
Detect changed workspaces

cristiano-belloni:
This PR adds a `getChangedWorkspaces` to `utils`, which is a function that takes a branch name and:

1. Gets all the files that git marks as changed between the current state of the repo and the latest commit in common with the specified branch, including untracked and unstaged files. This is handy, for example, to find out the files changed in a PR where the branch specified is the target branch.
1. Gets all the nearest `package.json` for each of the changed files to compute the containing workspace of each changed file.
1. Matches the list of workspace to the global `WorkspaceContent` and returns a subset `WorkspaceContent` containing only the changed workspace info.

## Example

```ts
const workspaces = await getAllWorkspaces();
const changedWorkspaces = await getChangedWorkspaces(workspaces, "origin/main");

// Assuming that only files in `workspace-1` and `workspace-3` have changed, the result would be:

[
  Map(
    {
      'workspace-1': {
        // Noisy package info, omissis...
      },
      'workspace-3': {
        // Noisy package info, omissis...
      },
    },
   ),
  {
    'workspace-1': {
      location: 'packages/workspace-1',
      workspaceDependencies: ['workspace-2', 'workspace-3'],
      mismatchedWorkspaceDependencies: [],
    },
    'workspace-3': {
      location: 'packages/workspace-3',
      workspaceDependencies: [],
      mismatchedWorkspaceDependencies: [],
    },
  },
]
```

## Tasks:

- [x] Match changed files to the closest package.json, which in turn matches to an entry in `workspaceInfo`
- [x] Return a subset of the workspaceInfo workspaces, with the same type
- [x] Make it possible to specify a branch in `getGitDiff`
- [x] Take workspaces in the arguments

### DIFF
--- a/packages/modular-scripts/src/utils/getChangedWorkspaces.ts
+++ b/packages/modular-scripts/src/utils/getChangedWorkspaces.ts
@@ -1,0 +1,30 @@
+import path from 'path';
+import { getDiffedFiles } from './gitActions';
+import { getAllWorkspaces } from './getAllWorkspaces';
+
+export async function getChangedWorkspaces(): Promise<Set<string>> {
+  const diffedFiles = getDiffedFiles();
+  console.log(diffedFiles);
+  const workspaces = await getAllWorkspaces();
+  const workspaceEntries = Object.entries(workspaces);
+  const changedWorkspaces: Set<string> = new Set();
+
+  for (const changedFile of diffedFiles) {
+    const changedWorkspace = workspaceEntries.find(([_, { location }]) =>
+      isSameOrSubdir(location, changedFile),
+    );
+    if (!changedWorkspace) {
+      // If the file changed is not in a workspace, all workspaces are marked as "changed"
+      return new Set(Object.keys(workspaces));
+    }
+    changedWorkspaces.add(changedWorkspace[0]);
+  }
+  return changedWorkspaces;
+}
+
+function isSameOrSubdir(parent: string, dir: string): boolean {
+  const relative = path.relative(parent, dir);
+  return (
+    !relative || (!relative.startsWith('..') && !path.isAbsolute(relative))
+  );
+}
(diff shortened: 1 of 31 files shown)


packages/modular-scripts/src/utils/gitActions.ts
```diff
@@ -55,10 +55,9 @@ function getModifiedAndUntrackedFileChanges(): string[] {
 }
 
 // Get all diffed files from git remote origin HEAD
-function getGitDiff(): string[] {
-  const defaultBranch = getGitDefaultBranch();
+function getGitDiff(targetBranch: string): string[] {
```
sgb-io, line 58: Is this def a backwards compatible change btw? I can see that the lint task relies on it

cristiano-belloni: This function is not exported - it's only internal and used in `getDiffedFiles`, which now has a default argument (so the lint task can invoke it as it did before)

sgb-io (approved):

packages/modular-scripts/src/utils/getChangedWorkspaces.ts
```diff
@@ -0,0 +1,65 @@
...
+  const [packages, workspaces] = workspaceContent;
+  const workspaceEntries = Object.entries(workspaces);
+  const result: WorkspaceContent = [new Map([]), {}];
+
+  for (const packagePath of packagePaths) {
+    const [resultPackages, resultWorkspaces] = result;
+    // Ignore holes
+    if (!packagePath) continue;
+    // Get the package directory from the package.json path and make it relative to the root, for comparison with the original WorkspaceContent
+    const packageDir = path.relative(root, path.dirname(packagePath));
```
steveukx, line 38: Should this check that the `packageDir` is a child of the `root` to prevent the possibility of finding `package.json` files outside of the repo

cristiano-belloni: I don't think it's possible; all the changed files that we find are inside the repo by definition (as we use the `git` command to find them), so we will _at least_ find the root `package.json`. If there's no root manifest, then the repo is not a modular repo and is not a monorepo, so we will never get here.

steveukx: 👍

PR #2027, opened 2022-08-11
Ghost testing

cristiano-belloni:
# Ghost testing

This PR implements three new options for the `test` command, which:

- Enable testing only changed workspaces, where "changed" means "containing files that are flagged as changed by Git, compared to a certain branch"
- Enable testing only changed workspaces as above, plus all their ancestors (i.e. the workspaces that directly or indirectly depend on them)
- Allow the user to specify the comparison branch (by default it's the default Git branch)

### `--ancestors`

Default: `false`

Can be used only in combination with `changed` or the command will fail. If set,
it will additionally execute tests for all the workspaces that (directly or
indirectly) depend on the workspaces selected by `changed`.

### `--changed`

Default: `false`

Execute tests only for the workspaces that contain files that have changed.
Files that have changed are calculated comparing the current state of the
repository with the branch specified by `compareBranch` or, if `compareBranch`
is not set, with the default git branch.

### `--compareBranch`

Default: `undefined`

Specify the comparison branch used to determine which files have changed when
using the `changed` option. If this option is used without `changed`, the
command will fail.

## TODO:
- [x] `--changed <branch>` testing only changed workspaces
- [x]  `--changed` boolean separated from `--compareBranch <branch>`
- [x]  `--ancestors` testing only changed workspaces and their ancestors
- [x] Option validation: `--ancestors` only with `--changed`
- [x] Option validation: `--changed` without regexes
- [x] High-level API that takes `WorkspaceContent`s and return `WorkspaceContent` with ancestors
- [x] modular-scripts docs
- [x] workspace-resolver docs (add high level API)
- [x] Unit tests
- [x] Integration tests


packages/modular-scripts/src/program.ts
```diff
@@ -155,6 +164,19 @@ program
...
+      process.exit(1);
+    }
+    if (options.changed && regexes.length) {
+      process.stderr.write(
+        'Option --changed conflicts with supplied test regex',
+      );
+      process.exit(1);
+    }
+
     const { default: test } = await import('./test');
```
steveukx, line 180: Does importing `./test` have side-effects?

cristiano-belloni: nope

packages/modular-scripts/src/program.ts
```diff
@@ -124,11 +124,20 @@ interface CLITestOptions extends TestOptions {
...
+    'Additionally run tests for workspaces that depend on workspaces that have changed',
+    false,
+  )
   .option(
     '--debug',
     'Setup node.js debugger on the test process - equivalent of setting --inspect-brk on a node.js process',
     false,
   )
+  .option(
+    '--changed <branch>',
```
steveukx, line 138: To allow for the future state of having `--changed` on its own to use the default / PR head, should this be a `boolean` with another option of `--compare-branch`

cristiano-belloni: Agree, it'd be clearer.

packages/modular-scripts/src/program.ts
steveukx: In that case it doesn't need a dynamic import - not something that's part of your PR, but might be easier to follow as a standard import at the top of the file.

cristiano-belloni: @steveukx in `program.ts` we parse the command line args and we decide which Modular command to execute. If we statically `import`ed all of them at the start (which is translated to sync `require` in cjs world), it'd be a performance hit proportional to the quantity of commands.

packages/modular-scripts/src/program.ts

packages/modular-scripts/src/__tests__/test.test.ts
```diff
@@ -28,45 +28,191 @@ function clearTests() {
...
+
+      fs.appendFileSync(
+        path.join(randomOutputFolder, '/packages/b/src/index.ts'),
+        "\n// Comment to package b's source",
+      );
+      fs.appendFileSync(
+        path.join(randomOutputFolder, '/packages/c/src/index.ts'),
+        "\n// Comment to package c's source",
+      );
+
```
sgb-io, line 147: Nice, funky tests :) Just a sense check: this will also work if the changes have already been committed in git, right?

sgb-io (approved):
LGTM

packages/modular-scripts/src/__tests__/test.test.ts
cristiano-belloni: It depends; this will just compare your current branch with a specified branch; so yes, if you have committed in `your-branch` (but haven't pushed your commit) and you use `origin/your-branch` as comparison branch, the tests will run.

To frame this a bit better: by default, `--changed` compares with the default branch, so the most common case is that you're in a feature branch, you branched from the repo's default branch and you want to know what tests to execute in CI. In this case, tests will run for everything that's changed from the default branch, even if the PR branch has been committed and pushed. I didn't want this test to branch and commit for simplicity, so I'm just modifying files on the fly on the main branch and leave them uncommitted.

PR #2050, opened 2022-08-20
Copy LICENSE files when building packages

joshwooding:
Part of #2160

### DIFF
--- a/packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
+++ b/packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
@@ -1,0 +1,35 @@
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import globby from 'globby';
+import getPrefixedLogger from '../../utils/getPrefixedLogger';
+import getModularRoot from '../../utils/getModularRoot';
+
+const licenseGlob = 'LICEN@(C|S)E*';
+
+export async function maybeCopyRootLicense(
+  target: string,
+  targetOutputDirectory: string,
+) {
+  const logger = getPrefixedLogger(target);
+  const modularRoot = getModularRoot();
+  const matches = globby.sync(path.join(targetOutputDirectory, licenseGlob), {
+    cwd: modularRoot,
+    onlyFiles: true,
+  });
+  if (matches.length === 0) {
+    logger.debug(
+      `No license found in ${targetOutputDirectory}. Looking for root license.`,
+    );
+    const rootLicenses = globby.sync(path.join(modularRoot, licenseGlob), {
+      cwd: modularRoot,
+      onlyFiles: true,
+    });
+    if (rootLicenses.length > 0) {
+      rootLicenses.forEach((license) => {
+        const filename = path.basename(license);
+        logger.log(`Copying ${filename} found in ${modularRoot}`);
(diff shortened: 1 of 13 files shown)


cristiano-belloni:
@joshwooding sorry for the delay, will take a look at this shortly in the next days

packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function maybeCopyRootLicense(
+  target: string,
+  targetOutputDirectory: string,
+) {
+  const logger = getPrefixedLogger(target);
+  const modularRoot = getModularRoot();
+
+  const targetOutputLicense = path.join(targetOutputDirectory, 'LICENSE');
+  if (!fs.existsSync(targetOutputLicense)) {
```
benpryke, line 14: Sometimes people use `LICENSE.txt`, `LICENSE.md` etc. Maybe we should copy any file of the form `LICENSE.*`?

cristiano-belloni:
LGTM, I think this is ready to merge after addressing @benpryke's feedback.

packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
```diff
@@ -0,0 +1,35 @@
...
+  target: string,
+  targetOutputDirectory: string,
+) {
+  const logger = getPrefixedLogger(target);
+  const modularRoot = getModularRoot();
+  const matches = globby.sync(path.join(targetOutputDirectory, licenseGlob), {
+    cwd: modularRoot,
+    onlyFiles: true,
+  });
+  if (matches.length == 0) {
```
benpryke, line 19: nit: `===`

benpryke: If the target package contains a license file, would this have been copied over by this point?

packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
```diff
@@ -0,0 +1,35 @@
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import globby from 'globby';
+import getPrefixedLogger from '../../utils/getPrefixedLogger';
+import getModularRoot from '../../utils/getModularRoot';
+
+const licenseGlob = 'LICEN@(C|S)E*';
+
+export async function maybeCopyRootLicense(
+  target: string,
```
benpryke, line 10: Should we not check for a license file in the target directory first?

packages/modular-scripts/src/build/buildPackage/maybeCopyRootLicense.ts
joshwooding: Yeah, due to pack-list and the default files array containing LICENSE. It's worth adding that if a user is using a different license format they would have to add it to files manually.

joshwooding: Unless you think modular should try and detect the license in that case too?

joshwooding:
Are there any more changes needed here?

joshwooding:
@cristiano-belloni Do you know if this PR can be merged too

PR #2051, opened 2022-08-20
Add --staged to modular lint

cheemcheem:
Adds support for a `--staged` flag to modular lint.

This flag conflicts with `--all` and only lint checks files that have been staged with git.

If the `--fix` flag is provided, it will run lint fix on only the staged files and then re-stage changes on **only** the previously staged files.

This makes `modular lint --fix --staged` a good pre-commit hook, especially when paired with `eslint-plugin-prettier`.

Tested `--staged` with no changes, non-staged changes, partially staged changes, and fully staged changes on a repository with and without the `--fix` flag to ensure it functions correctly. I didn't see any place for unit tests for `gitActions.ts` so I didn't write any. The `lint.ts` tests also didn't test any lint flags so I also didn't test `--staged` there.

### DIFF
--- a/packages/modular-scripts/src/lint.ts
+++ b/packages/modular-scripts/src/lint.ts
@@ -6,28 +6,29 @@ import actionPreflightCheck from './utils/actionPreflightCheck';
 import { resolveAsBin } from './utils/resolveAsBin';
 import getModularRoot from './utils/getModularRoot';
 import execAsync from './utils/execAsync';
-import { getDiffedFiles } from './utils/gitActions';
+import { addFiles, getDiffedFiles, getStagedFiles } from './utils/gitActions';
 import * as logger from './utils/logger';
 
 export interface LintOptions {
   all: boolean;
   fix: boolean;
+  staged: boolean;
 }
 
 async function lint(
   options: LintOptions,
   regexes: string[] = [],
 ): Promise<void> {
-  const { all = false, fix = false } = options;
+  const { all = false, fix = false, staged = false } = options;
   const modularRoot = getModularRoot();
   const lintExtensions = ['.ts', '.tsx', '.js', '.jsx'];
   let targetedFiles = ['<rootDir>/**/src/**/*.{js,jsx,ts,tsx}'];
 
   if (!all && !isCI && regexes.length === 0) {
-    const diffedFiles = getDiffedFiles();
+    const diffedFiles = staged ? getStagedFiles() : getDiffedFiles();
     if (diffedFiles.length === 0) {
       logger.log(
-        'No diffed files detected. Add the `--all` option to lint the entire codebase',
+        'No diffed files detected. Use the `--all` option to lint the entire codebase',
       );
       return;
     }
@@ -71,6 +72,11 @@ async function lint(
         MODULAR_LINT_FIX: String(fix),
       },
     });
+
+    if (staged && fix) {
+      targetedFiles = targetedFiles.map((p) => p.replace('<rootDir>/', ''));
+      addFiles(targetedFiles);
+    }
   } catch (err) {
     logger.debug((err as ExecaError).message);
     // ✕ Modular lint did not pass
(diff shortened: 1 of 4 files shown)


packages/modular-scripts/src/lint.ts
```diff
@@ -6,28 +6,29 @@ import actionPreflightCheck from './utils/actionPreflightCheck';
...
   const lintExtensions = ['.ts', '.tsx', '.js', '.jsx'];
   let targetedFiles = ['<rootDir>/**/src/**/*.{js,jsx,ts,tsx}'];
 
   if (!all && !isCI && regexes.length === 0) {
-    const diffedFiles = getDiffedFiles();
+    const diffedFiles = staged ? getStagedFiles() : getDiffedFiles();
     if (diffedFiles.length === 0) {
       logger.log(
-        'No diffed files detected. Add the `--all` option to lint the entire codebase',
+        'No diffed files detected. Use the `--all` option to lint the entire codebase',
```
cheemcheem, line 31: Updated wording so it doesn't suggest '--all' can be used with '--staged'.

packages/modular-scripts/src/lint.ts
```diff
@@ -71,6 +72,11 @@ async function lint(
         MODULAR_LINT_FIX: String(fix),
       },
     });
+
+    if (staged && fix) {
+      targetedFiles = targetedFiles.map((p) => p.replace('<rootDir>/', ''));
```
cheemcheem, line 77: `<rootDir>` is added for eslint purposes but that is not required for git re-staging purposes.

packages/modular-scripts/src/program.ts
```diff
@@ -245,13 +245,23 @@ program
...
-  .option('--fix', 'Fix the lint errors wherever possible')
+  .option(
+    '--fix',
+    `Fix the lint errors wherever possible, restages changes if run with ${lintStagedFlag}`,
+  )
+  .addOption(
+    new Option(
+      lintStagedFlag,
+      'Only lint files that have been staged to be committed',
+    ).conflicts('all'),
```
cheemcheem, line 263: `--all` shouldn't be allowed at the same time as `--staged`.

packages/modular-scripts/src/utils/gitActions.ts
```diff
@@ -54,13 +54,8 @@ function getModifiedAndUntrackedFileChanges(): string[] {
...
 
-// Get all diffed files from git remote origin HEAD
-function getGitDiff(targetBranch: string): string[] {
-  // get a commit sha between the HEAD of the current branch and git remote origin HEAD
-  const sha = execa.sync('git', ['merge-base', 'HEAD', targetBranch], {
-    cwd: getModularRoot(),
-  });
-
+// Get all diffed files in standardised order exluding unknown and removed files
+function runGitDiff(gitDiffCommand: string): string[] {
```
cheemcheem, line 58: Extracted this functionality out to its own function so it can be used from both `getGitDiff` and `getStagedFiles`.

packages/modular-scripts/src/program.ts
```diff
@@ -245,13 +245,23 @@ program
...
     'Only lint diffed files from your remote origin default branch (e.g. main or master)',
   )
-  .option('--fix', 'Fix the lint errors wherever possible')
+  .option(
+    '--fix',
+    `Fix the lint errors wherever possible, restages changes if run with ${lintStagedFlag}`,
+  )
+  .addOption(
+    new Option(
+      lintStagedFlag,
```
sgb-io, line 261: Tiny nit, the other options are just static strings, so for consistency's sake:

```suggestion
      '--staged',
```

cheemcheem: I saw that, however the `--fix` help text in line 257 of the same file also refers to `--staged`, and I thought in future if the `--staged` flag were to change, a developer might not realise that the `--fix` help text also needs updating. So I created this variable and used it in both places.

It's not a big deal as the usages of `lintStagedFlag` are very close together, so if you think it would be better to maintain consistency then that is fine with me.

cristiano-belloni (approved):

PR #2130, opened 2022-09-12
chore: reduce dependabot frequency and ignore major version bumps

ayseenver:


cristiano-belloni (approved):

.github/dependabot.yml
```diff
@@ -8,12 +8,16 @@ updates:
   - package-ecosystem: 'npm' # See documentation for possible values
     directory: '/' # Location of package manifests
     schedule:
-      interval: 'daily'
-    # Allow up to 10 open pull requests for NPM dependencies
+      interval: 'weekly'
+    # Allow up to 100 open pull requests for NPM dependencies
     open-pull-requests-limit: 100
```
sgb-io, line 13: Should we reduce this? Currently, the amount of open PRs is quite noisy. I'd vote to cut this down, but I'm also unsure what the implications might be

PR #2142, opened 2022-09-14
chore: replace ts jest with swc

ayseenver:


package.json
```diff
@@ -115,5 +115,12 @@
       "yarn lint"
     ],
     "*": "node scripts/validate-lockfile.js"
+  },
+  "devDependencies": {
+    "@swc/core": "^1.3.0",
+    "@swc/jest": "^0.2.21",
+    "@types/lodash": "^4.14.185",
```
ayseenver, line 122: I had to add these lodash types to be able to commit

package.json
```diff
@@ -115,5 +115,12 @@
       "yarn lint"
     ],
     "*": "node scripts/validate-lockfile.js"
+  },
+  "devDependencies": {
```
cristiano-belloni, line 119: Why in the root package.json? Shouldn't it be the modular-scripts one?

PR #2144, opened 2022-09-14
Create test for current state of the `verifyWorkspaceDependencies` check

steveukx:
Codifies some of the current behaviour of the `verifyWorkspaceDependencies` check to simplify making changes to it in the future.

### DIFF
--- a/__fixtures__/verify-workspace/package.json
+++ b/__fixtures__/verify-workspace/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "clean-workspace-3",
+  "version": "1.0.0",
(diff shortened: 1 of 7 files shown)


__fixtures__/verify-workspace/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "clean-workspace-3",
```
cristiano-belloni, line 2: I think it's bit confusing using "workspace" here, since workspaces are the children packages, not the root one. Maybe "clean-modular-project-3" would be better?

cristiano-belloni (approved):

PR #2145, opened 2022-09-14
Selective test on workspace name with the --package option

cristiano-belloni:
Enhance `modular test` with the following behaviour:
```
$ modular test --package PACKAGE_NAME
# Run all tests for PACKAGE_NAME

$ modular test --package PACKAGE_NAME_1 --package PACKAGE_NAME_2
# Run all tests for PACKAGE_NAME_1 and PACKAGE_NAME_2

$ modular test --package PACKAGE_NAME --ancestors
# Run all tests for PACKAGE_NAME and all other workspaces that, directly or indirectly, depend on it

$ modular test --package PACKAGE_NAME --changed
# Options are not compatible: exit with error

modular test SOME_REGEX --package PACKAGE_NAME
# Options are not compatible: exit with error
```

### DIFF
--- a/docs/commands/test.md
+++ b/docs/commands/test.md
@@ -64,7 +64,7 @@ Specify the comparison branch used to determine which files have changed when
 using the `changed` option. If this option is used without `changed`, the
 command will fail.
 
-#### collectCoverageFrom
+### collectCoverageFrom
 
 [_Documentation_](https://jestjs.io/docs/configuration#collectcoveragefrom-array)
 
@@ -88,7 +88,7 @@ Example:
 }
 ```
 
-#### coveragePathIgnorePatterns
+### coveragePathIgnorePatterns
 
 [_Documentation_](https://jestjs.io/docs/configuration#coveragepathignorepatterns-arraystring)
 
@@ -98,7 +98,7 @@ An array of regexp pattern strings that are matched against all file paths
 before executing the test. If the file path matches any of the patterns,
 coverage information will be skipped.
 
-#### coverageThreshold
+### coverageThreshold
 
 [_Documentation_](https://jestjs.io/docs/configuration#coveragethreshold-object)
 
@@ -111,7 +111,7 @@ positive number are taken to be the minimum percentage required. Thresholds
 specified as a negative number represent the maximum number of uncovered
 entities allowed.
 
-#### moduleNameMapper
+### moduleNameMapper
 
 [_Documentation_](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring)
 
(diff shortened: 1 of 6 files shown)


docs/commands/test.md
```diff
@@ -142,7 +142,16 @@ before those paths are to be considered 'visible' to the module loader. If a
 given module's path matches any of the patterns, it will not be `require()`-able
 in the test environment.
 
-#### testPathIgnorePatterns
+### package
+
+Default: `undefined`
+
+Run all the test for the workspace with the specified package name. Can be
```
sgb-io, line 149: Small typo (I think?)

```suggestion
Run all the tests for the workspace with the specified package name. Can be
```

docs/commands/test.md
```diff
@@ -142,7 +142,16 @@ before those paths are to be considered 'visible' to the module loader. If a
...
+repeated to select more than one workspace. Can be combined with the
```
sgb-io, line 150: Also a typo?

```suggestion
repeated to select more than one package. Can be combined with the
```

sgb-io: Maybe not. I'm confused - are they workspaces or packages? In my mind, a workspace can consist of many packages (the workspace usually means the `packages` dir, if talking in yarn languge)

packages/modular-scripts/src/__tests__/test.test.ts
```diff
@@ -194,6 +194,209 @@ describe('Modular test command', () => {
...
+      expect(resultPackages.stderr).toContain(
+        'packages/b/src/__tests__/b.test.ts',
+      );
+
+      const resultPackagesWithAncestors = runRemoteModularTest(
+        currentModularFolder,
+        randomOutputFolder,
+        ['test', '--ancestors', '--package', 'b', '--package', 'c'],
+      );
+      expect(resultPackagesWithAncestors.stderr).toContain(
```
sgb-io, line 246: nit: could maybe use a single `expect` statement here

cristiano-belloni: Unless you're saying that we can conflate `packages/c/src/__tests__/utils/c-nested.test.ts` and `packages/c/src/__tests__/utils/c-nested.test.ts` into their common root: `packages/c/src/__tests__/`; in that case I explicitly wanted to show that we're executing all the tests of a workspace, including those in subdirectories.

sgb-io: What's the shape of `resultPackagesWithAncestors.stderr`? I can imagine we can find a way to dedupe this a bit, but it's really not a big deal

sgb-io (approved):

PR #2146, opened 2022-09-16
Add support for extendedPublishConfig, apply to workspace resolver

sgb-io:
Thanks @steveukx for providing prior art in [git-js](https://github.com/steveukx/git-js/blob/main/simple-git/scripts/package-json.js)! My implementation is very similar, but I've come up with some alternative naming, let me know what you think.

- Adds a script to the root which enables a `extendedPublishConfig` block in the `package.json` of any packages in the project. The idea is that we can make changes to the project before publishing, very similar to the native [publishConfig](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#publishconfig) option, however, is not restricted to what fields we can change.
- All items in `extendedPublishConfig` will override fields of the same name at publish time
- We also remove `scripts`, since they are useless in published packages. This isn't required to achieve the desired result here, but could help avoid confusion should a user attempt to use scripts in the wrong way
- Applied `extendedPublishConfig` to the workspace resolver package. In this case, we have made the `main` and `types` fields change to the built versions at publish time.

The end result is that developers no longer need to build the workspace resolver package in local dev - the dependency can be imported right away, and the types just work.

### DIFF
--- a/scripts/extend-publish-config.js
+++ b/scripts/extend-publish-config.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const { writeFileSync } = require('fs');
+const { resolve } = require('path');
+
+// This script applies the contents of any `extendedPublishConfig` blocks
+// to the root package.json of affected packages in PACKAGES_TO_EXTEND.
+//
+// It also removes `scripts` on the assumption none are needed when installing npm libraries.
+
+const PACKAGES_TO_EXTEND = ['workspace-resolver'];
+
+function writeNewPackageJson(src, content) {
+  writeFileSync(src, JSON.stringify(content, null, 2));
+}
+
+function updatePackageJson(src) {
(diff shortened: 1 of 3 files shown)


scripts/extend-publish-config.js
```diff
@@ -0,0 +1,30 @@
...
+
+const { writeFileSync } = require('fs');
+const { resolve } = require('path');
+
+// This script applies the contents of any `extendedPublishConfig` blocks
+// to the root package.json of affected packages in PACKAGES_TO_EXTEND.
+//
+// It also removes `scripts` on the assumption none are needed when installing npm libraries.
+
+const PACKAGES_TO_EXTEND = ['workspace-resolver'];
```
sgb-io, line 11: Technically this is directory names. I figured that having a hardcoded static list of packages to apply this to, is fine for the scope of this project.

cristiano-belloni (approved):

PR #2167, opened 2022-09-27
Support for non-modular packages in tests and workspace resolver

cristiano-belloni:
This PR stems from the discussion here: https://github.com/jpmorganchase/modular/discussions/2168 and from the necessity of either:
a) not list non-modular workspaces in workspace-resolver or 
b) list them, but clarifying that they are non-modular.

Since we don't want to impact how `mismatchedWorkspaceDependencies` works and we like the idea of executing workspace commands on non-modular workspaces, this PR modifies the output of workspace-resolver, adding a `type` field, and adds support for non-modular (from now: extraneous) workspaces in `modular test`. In particular, `modular test` will:

- identify the eventual extraneous packages with selective options (`--package`, `--changed`, `--ancestors`), in addition to the usual modular packages
- try to run `yarn workspace <extraneous-workspace-name> test` for each one of them, **after** running the regular jest tests
- emit a warning for all the selected extraneous packages that don't have a test script
- fail the test command if one of them fails
- succeed normally if all of them succeed

I also refactored the `test` command to factor out utilities and tasks.

Note: The PR seems huge, but most of the content are new test fixtures to simulate a monorepo with examples of failing, successful and non-existent test scripts in extraneous workspaces, with regular Modular workspace ancestors.

TASKS

- [x] `type` is first-class
- [x] Run package.json `test` script for non-modular workspaces or warn

### DIFF
--- a/packages/workspace-resolver/src/resolve-workspace.ts
+++ b/packages/workspace-resolver/src/resolve-workspace.ts
@@ -65,7 +65,8 @@ export async function resolveWorkspace(
   const pkgPath = packageJsonPath(root);
 
   const json = await readPackageJson(isRoot, workingDirToUse, pkgPath);
-  const isModularRoot = json.modular?.type === 'root';
+  const type = json.modular?.type || undefined;
+  const isModularRoot = type === 'root';
 
   if (!json.name) {
     throw new Error(
@@ -89,10 +90,8 @@ export async function resolveWorkspace(
     workspace: !!json.workspaces,
     children: [],
     parent,
-    modular: {
-      type: 'unknown',
-      ...json.modular,
-    },
+    modular: json.modular,
+    type,
     // Like yarn classic `workspaces info`, we include all except peerDependencies
     dependencies: {
       ...json.optionalDependencies,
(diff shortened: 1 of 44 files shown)


sgb-io (commented):
LGTM, although I would suggest we update the `README.md` for workspace-resolver, so that this change is publicly documented (mainly the existence of the new `type` field and it's purpose).

packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -153,7 +152,7 @@ export function analyzeWorkspaceDependencies(
   // Exclude the root when analyzing package inter-dependencies
   const packagesWithoutRoot = Array.from(workspacePackages.entries()).filter(
     ([, pkg]) => {
-      return pkg.modular.type !== 'root';
+      return pkg.modular?.type !== 'root';
```
steveukx, line 155: Minor thing, but this filter can be inlined to the next `forEach` on line 160 to avoid looping through the workspace twice.

As `type` has been added to the `ModularWorkspacePackage`, this can now be:

```suggestion
      return pkg.type !== 'root';
```

packages/workspace-resolver/src/resolve-workspace.ts
```diff
@@ -65,7 +65,8 @@ export async function resolveWorkspace(
   const pkgPath = packageJsonPath(root);
 
   const json = await readPackageJson(isRoot, workingDirToUse, pkgPath);
-  const isModularRoot = json.modular?.type === 'root';
+  const type = json.modular?.type || undefined;
```
steveukx, line 68: Do we need to cater for converting to `undefined` any `modular.type` that has been explicitly set to `null`?

```suggestion
  const type = json.modular?.type;
```

cristiano-belloni:
Closing as it needs to be discussed further

PR #2169, opened 2022-09-27
Allow workspaces outside the packages/ directory

cristiano-belloni:
The verifyWorkspaceDependencies check should not reject modular packages found outside of the packages directory.

Consider the file system:

```
- package.json
- apps
  |- some-app-not-in-packages
- packages
  |- package-a
  |- package-b (not a modular package)
  ```
When the modular check runs as part of pre-flight, the some-app-not-in-packages package will cause the check to fail because it is outside the packages directory.

Given that it was found by the workspace-resolver, the check should not fail.

### DIFF
--- a/__fixtures__/verifiable-project/packages/app-one/package.json
+++ b/__fixtures__/verifiable-project/packages/app-one/package.json
@@ -6,9 +6,9 @@
   },
   "dependencies": {
     "package-one": "workspace:*",
-    "package-two": "workspace:^",
-    "package-three": "workspace:~",
-    "package-four": "workspace:^1.0.0"
+    "package-two": "workspace:*",
+    "package-three": "workspace:*",
+    "package-four": "workspace:*"
   },
   "version": "1.0.0"
 }
--- a/packages/modular-scripts/src/__tests__/check/verifyWorkspaceDependencies.test.ts
+++ b/packages/modular-scripts/src/__tests__/check/verifyWorkspaceDependencies.test.ts
@@ -22,16 +22,12 @@ describe('verifyWorkspaceDependencies', () => {
     expect(error).not.toHaveBeenCalled();
   });
 
-  it('rejects packages not in the "packages" directory', async () => {
+  it('doesn\'t reject packages not in the "packages" directory', async () => {
     mock(getModularRoot).mockReturnValue('__fixtures__/verifiable-project');
 
     const checked = await check('__fixtures__/verifiable-project');
-    expect(checked).toBe(false);
-    expect(error).toHaveBeenCalledWith(
-      expect.stringContaining(
-        `package-four is not located within the "/packages" directory`,
-      ),
-    );
+    expect(checked).toBe(true);
+    expect(error).not.toHaveBeenCalled();
   });
 });
 
--- a/packages/modular-scripts/src/check/verifyWorkspaceDependencies.ts
+++ b/packages/modular-scripts/src/check/verifyWorkspaceDependencies.ts
@@ -29,13 +29,6 @@ export async function check(target?: string): Promise<boolean> {
       valid = false;
     }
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: __fixtures__/verifiable-project/packages/app-one/package.json [hunk 1]
```diff
@@ -6,9 +6,9 @@
   },
   "dependencies": {
     "package-one": "workspace:*",
-    "package-two": "workspace:^",
```
sgb-io, line 9: Did you need to change this? I think by having a vareity of semver syntax, it helps increase confidence in our semver checking logic

packages/modular-scripts/src/check/verifyWorkspaceDependencies.ts
```diff
@@ -29,13 +29,6 @@ export async function check(target?: string): Promise<boolean> {
       valid = false;
     }
 
-    if (!packageInfo.location.startsWith('packages')) {
```
sgb-io, line 32: Shouldn't we at least limit it to packages found in the `workspaces` field in `package.json`? This would keep a level of compatibility with yarn. Yarn implements that boundary.

Won't it potentially pick up stuff in e.g. `__tests__` with zero restriction?

packages/modular-scripts/src/__tests__/check/verifyWorkspaceDependencies.test.ts
```diff
@@ -22,16 +22,12 @@ describe('verifyWorkspaceDependencies', () => {
     expect(error).not.toHaveBeenCalled();
   });
 
-  it('rejects packages not in the "packages" directory', async () => {
+  it('doesn\'t reject packages not in the "packages" directory', async () => {
```
sgb-io, line 25: I think we need a new test case like the one given in your PR description

__fixtures__/verifiable-project/packages/app-one/package.json
cristiano-belloni: This is another test fixture, specifically made to `verifyWorkspaceDependencies`.
I think @steveukx just copy-pasted the original fixture, `clean-workspace-3`, that you created to test mismatched dependencies.
In the world pre-this PR, `verifyWorkspaceDependencies` errored straight away because one of the workspaces was outside `packages/`.
In this PR, we want to assert that `verifyWorkspaceDependencies` doesn't error if one of the workspaces is outside `packages/`. The problem is that now the check proceeds, finds mismatched dependencies and errors anyway. That's why I need all dependencies being matched in this specific case.

packages/modular-scripts/src/check/verifyWorkspaceDependencies.ts
cristiano-belloni: I don't think so; we get our workspaces through `getWorkspaceInfo`, which in turn uses the workspace resolver to scan all the workspaces in the repo, looking at the `workspaces` field in `package.json`. As a consequence, we can't get a workspace outside `workspaces`, since we use `workspaces` in the first place to generate their list.

packages/modular-scripts/src/__tests__/check/verifyWorkspaceDependencies.test.ts
cristiano-belloni: Not sure what you mean here.

__fixtures__/verifiable-project/packages/app-one/package.json
sgb-io: Understood 👍

packages/modular-scripts/src/__tests__/check/verifyWorkspaceDependencies.test.ts
sgb-io: I think given your other comments, we don't need a new test case, you can disregard this

sgb-io (approved):

PR #2186, opened 2022-10-19
CSS from CDN to not be injected in `<head>`, but listed in package.json + synthetic index.html

cristiano-belloni:
**Before this PR**: Non-local CSS `import`ed in the code is rewritten to CDN at build time and injected in the page `<head>` at runtime.

**After this PR**: Non-local CSS `import`ed in the code is rewritten to CDN and included in the synthetic index's `<head>` at build time + listed in the output manifest.

**Result**: Non-local CSS is still loaded seamlessly when the esm-view is served standalone, but not loaded when the esm-view is dynamically imported at runtime. This allows, for example, an host application to look at an esm-view's output manifest before loading it and decide whether to load its non-local CSS (which might be already present in the browser page).

## TODO

- [x] webpack loader not injecting CSS from CDN
- [x] esbuild plugin not injecting CSS from CDN
- [x] extract CSS imports from source
- [x] insert css CDN URLs in ~`style-cdn`~ `styleImports` field in output package.json
- [x] insert css CDN URLs in generated synthetic HTML ~`<style>`~ `<link>` tags (build)
- [x] insert css CDN URLs in generated synthetic HTML ~`<style>`~ `<link>` tags (start)
- [x] check that bundled css is still bundled
- [x] check behaviour with CSS imported from .css files vs CSS imported via ts import (if they diverge, document it)
- [x] docs
- [x] tests


packages/modular-scripts/src/build/index.ts
```diff
@@ -258,21 +238,21 @@ async function buildStandalone(
...
       modular: targetPackageJson.modular,
-      dependencies: targetPackageJson.dependencies,
-      bundledDependencies: targetPackageJson.bundledDependencies,
+      dependencies: packageDependencies,
+      bundledDependencies: isApp
+        ? Object.keys(packageDependencies)
+        : Object.keys(bundledResolutions),
       module: jsEntryPoint ? paths.publicUrlOrPath + jsEntryPoint : undefined,
       style: cssEntryPoint ? paths.publicUrlOrPath + cssEntryPoint : undefined,
+      'style-cdn': styleImports?.size ? [...styleImports] : undefined,
```
steveukx, line 255: Could this follow the naming convention of `camelCase` rather than `kebab-case` for the property?

As these are `styleImports` elsewhere, might that be a useful name here?

packages/modular-scripts/src/esbuild-scripts/api.ts
```diff
@@ -153,6 +161,16 @@ function compileIndex({
     (node) => node.nodeName === 'body',
   ) as parse5.Element;
 
+  if (styleImports) {
+    [...styleImports].forEach((importUrl) =>
+      head.childNodes.push(
+        ...parse5.parseFragment(
+          `<link rel="stylesheet" href="${importUrl}"></script>`,
```
steveukx, line 168: Might this be simpler to read with `insertAdjacentHTML`

```typescript
styleImports?.length && head.insertAdjacentHtml('beforeEnd', 
  styleImports.map(importUrl => `<link rel="stylesheet" href="${importUrl}" />`).join('')
);
```

Note - there is a typo in the template being used - it opens a `link` and closes a `script` 

```suggestion
          `<link rel="stylesheet" href="${importUrl}" />`,
```

cristiano-belloni: > Might this be simpler to read with insertAdjacentHTML

`parse5.Element` instances do not seem to support `insertAdjacentHTML`. I think parse5's API is reduced compared to the real DOM API.

> Note - there is a typo in the template being used - it opens a link and closes a script

Lol, a copy paste went wrong. Thanks for spotting it.

packages/modular-scripts/src/build/index.ts
cristiano-belloni: Set it to `styleImports`, it seems that nobody else's using it, so we can claim it in this npm Wild West :D

sgb-io (approved):

PR #2190, opened 2022-10-24
Fix modular add from custom template

AlbertoBrusa:
- Fixed modular add from template not copying all files from the template as intended when no filter "files" array is specified in the template's package.json
- Added additional logging for clarity 
- Added details to Templates documentation


docs/concepts/templates.md
```diff
@@ -13,6 +13,11 @@ either used implicitly (every modular `type` has a default template) or provided
 explicitly to `modular add`, either interactively or via the `--template`
 option.
 
+Templates can either be defined and used locally within your workplace, or
```
cristiano-belloni, line 16: Typo: "workspace"

packages/modular-scripts/src/addPackage.ts
```diff
@@ -203,22 +206,24 @@ async function addPackage({
...
 
-  const packageFilePaths = getAllFiles(packagePath);
+  const packageFilePaths = getAllFiles(newPackagePath);
 
   // If we get our package locally we need to allowlist files like yarn publish does
+  // If package.json specifies "files" to include, filter, otherwise allow all files
   const packageAllowlist = globby
-    .sync(modularTemplatePackageJson.files || ['*'], {
-      cwd: packagePath,
+    .sync(modularTemplatePackageJson.files || ['**'], {
```
cristiano-belloni, line 225: Maybe a bit out of scope, but it turns out there's a npm package to do just this the "official" way, instead of the globby hack we're using here: https://github.com/npm/npm-packlist

Since we're improving local `files` filtering anyway, what do you think if we just use this package instead?

AlbertoBrusa: Switched to packlist with my latest changes - surprised that it requires @npmcli/arborist and added over 50 indirect dependencies to Modular, seems excessive for such a small bit of functionality

sgb-io: That is a bit suprising. Taking a look, I guess it makes sense why: `Arborist` is npm's internal representation of `node_modules`. The [arborist library](https://github.com/npm/cli/tree/latest/workspaces/arborist) (TIL it exists) is a bit similar to our workspace resolver package (but different outcomes - eg building an ideal tree whereas we're interested in ancestors and other things). So when we add `npm-packlist` we get all of that as devDeps.

Assuming that we're working within the bounds of npm/yarn (yarn being npm-compatible) (not sure about pnpm but my assumption would be that it's also npm-compatible) for a long time (ignoring things like Deno and Bun that seem definitely out of scope), we get the guarantee of being spec/design compliant, less can go wrong. So there is some value for sure.

An example is the "common cruft files" described in that package's readme ([see rule number 5](https://github.com/npm/npm-packlist)) - I doubt `**` results in full/proper compliance vs using `npm-packlist`.

AlbertoBrusa: Thanks, that's some valuable insight and some really good points. I see the value in using it but I'm seeing it generate some curious errors in the tests which I'll investigate tomorrow.

packages/modular-scripts/src/addPackage.ts
```diff
@@ -181,11 +187,13 @@ async function addPackage({
     }
 
     await templateInstallSubprocess;
-    newModularPackageJsonPath = require.resolve(installedPackageJsonPath);
+    templatePackageJsonPath = require.resolve(installedPackageJsonPath, {
+      paths: [modularRoot],
```
AlbertoBrusa, line 191: When calling modular scripts from another modular repository it Require.resolve would resolve in the repo holding the script, instead of what it's run on.

Changed require.resolve() to always check from the Modular Root as identified by modular, which is the repo modular is run on, not from.

Curious to see if anyone has any concerns about this change, I feel like it's actually an improvement that will provide more consistent behaviour, and only really affect tests/us devs when we want to test modular changes.

packages/modular-scripts/src/__tests__/addPackage.test.ts
```diff
@@ -2,91 +2,84 @@ import execa from 'execa';
...
     ).rejects.toThrow();
   });
 });
 
 describe('When working with a scoped app', () => {
   beforeAll(async () => {
-    await modular('add @scoped/sample-app --unstable-type app', {
-      stdio: 'inherit',
-    });
+    await modular('add @scoped/sample-app --unstable-type app');
```
steveukx, line 54: As this is in a `beforeAll` rather than `beforeEach` the tests will all use the same temp directory, meaning they could depend on the state left by previous tests which makes it less clear what is being tested.

As the simplest option, `createModularTestContext` in a `beforeEach` to set up the test directory for each test. If you find that the tests are taking too long you could create a context, add the app to it then clone that directory in a `beforeEach`

AlbertoBrusa: That was my intention for performance reasons, but I agree beforeEach would provide a more 'solid'/isolated test context - I'll check the performance impact and explore your cloning suggestion, thanks

packages/modular-scripts/src/check/verifyPackageTree.ts
```diff
@@ -47,17 +48,24 @@ export async function check(): Promise<boolean> {
...
   );
-
+  let logOutCount = 0;
   for (const dep of DEPS_TO_CHECK) {
     let currentDir = startDirectory;
 
     // We've reached the root of the modular repo
     // in which case we want to bail out
-    while (currentDir !== path.resolve('/')) {
+    // Stop if we've reached the drive's root
```
AlbertoBrusa, line 61: With this change I was able to stop the infinite loop and get the tests to pass - I'm assuming because of the symlinked node modules it was getting to root as it never reached modular root, as 
`equire.resolve('modular-scripts/package.json', {
      paths: [getModularRoot()],
    }),`
would resolve to modular/node_modules, not the temp  modular repo's node_modules 

@cristiano-belloni let me know if this change seems sound to you of if you have any concerns

packages/modular-scripts/src/__tests__/addPackage.test.ts
```diff
@@ -2,91 +2,84 @@ import execa from 'execa';
...
+ */
+function cleanup() {
+  rimraf.sync(packagesPath);
+  fs.mkdirSync(packagesPath);
+}
 
 function modular(str: string, opts: Record<string, unknown> = {}) {
-  return execa('yarnpkg', ['modular', ...str.split(' ')], {
-    cwd: modularRoot,
+  return execa('yarnpkg', ['modular', ...str.split(' '), '--verbose'], {
```
cristiano-belloni, line 34: Commenting to remind us to remove this at clean-up

AlbertoBrusa: Thanks!

packages/modular-scripts/src/addPackage.ts
cristiano-belloni: I don't see any evident downside to this. 👍

packages/modular-scripts/src/check/verifyPackageTree.ts
cristiano-belloni: I love that resolving `..` of root yields root itself instead of throwing, that's some kind of unmovable motor paradox that launches our checks into infinity.

I don't see any downside in the extra check since nothing can be parent of root (but itself) and I hope users don't host their Modular projects on `/`; it might even harden us against Windows junction oddities, so good for me.

packages/modular-scripts/src/test/utils.ts
```diff
@@ -53,3 +56,74 @@ export async function addFixturePackage(
...
+ */
+export function createModularTestContext(): string {
+  // Modular node_modules are copied in the parent folder
+  const tempDir = tmp.dirSync().name;
+  fs.symlinkSync(
+    path.join(modularRoot, 'node_modules'),
+    path.join(tempDir, 'node_modules'),
+  );
+
+  // Actual mock modular repo is nested underneath so that it can have it's own node_modules
```
cristiano-belloni, line 72: its

packages/modular-scripts/src/test/utils.ts
```diff
@@ -53,3 +56,74 @@ export async function addFixturePackage(
     );
   }
 }
+
+/**
+ * Creates a temporary Modular repo with Modular node_modules symlinked
```
cristiano-belloni, line 61: Can you describe a bit better how this works + the motivations behind this for future memory? I expect people reading this code not understanding why we're doing this immediately.

AlbertoBrusa: Yep, will add more details

packages/modular-scripts/src/__tests__/addPackage.test.ts
AlbertoBrusa: Performance difference setting up the directory for each test was negligible so went ahead and made the change 
The test context is generated in the BeforeAll for each test as the tests were written in such a way that expects the results of previous assertion to still be there, see:

```
describe('When working with a scoped app', () => {
  beforeAll(async () => {
    createTempModularRepoWithTemplate(appTemplatePath);
    await modular('add @scoped/sample-app --unstable-type app');
  });

  it('creates the app in the expected directory, with the expected name', async () => {
    const manifest = (await fs.readJSON(
      path.join(tempPackagesPath, 'sample-app', 'package.json'),
    )) as CoreProperties;
    expect(manifest.name).toEqual('@scoped/sample-app');
  });

  it('fails if trying to add another app with the same name', async () => {
    await expect(
      modular('add @scoped/sample-app --unstable-type app'),
    ).rejects.toThrow();
  });
```

packages/modular-scripts/src/__tests__/addPackage.test.ts
AlbertoBrusa: Thanks, done

packages/modular-scripts/src/test/utils.ts
AlbertoBrusa: Added some detail, let me know if you think it's sufficient

packages/modular-scripts/src/test/utils.ts
```diff
@@ -53,3 +56,78 @@ export async function addFixturePackage(
...
+    spaces: 2,
+  });
+
+  return tempModularRepo;
+}
+
+/**
+ * Mocks the installation of a template in the provided modular repo to avoid running `yarn` or installing it from a registry.
+ *
+ * Copies the template into a Templates workspace and symlinks them in node_modules so that it can be picked up by moudlar add.
```
cristiano-belloni, line 111: moudlar -> modular

cristiano-belloni (approved):

PR #2191, opened 2022-10-24
Add convert-to-template command

AlbertoBrusa:
Added a command that converts a modular package into a template


AlbertoBrusa:
Closing as we want less commands to support not more :)

PR #2194, opened 2022-10-28
Feat/modular package builder

cristiano-belloni:
This draft PR has been created to discuss the plugin approach.

Topics to discuss:

- [ ] Abstract yarn commands to a separate package to allow support of pnpm (or other non-yarn) package manager


packages/build-standalone/src/plugins/esbuild-measure-file-size.ts
```diff
@@ -0,0 +1,95 @@
...
+export function esbuildMeasureFileSizes<T extends ModularBuildContext>(
+  config: ModularBuildConfig,
+): WorkflowPlugin<T>[] {
+  return [
+    {
+      phase: 'validate',
+      async handler(context: T) {
+        return {
+          ...context,
+          [FILE_SIZES]: await measureBeforeBuild(
```
cristiano-belloni, line 21: Every plugin enriches the plugin chain state by returning a new copy of the context. How do we handle cases in which one plugin is dependent on another? For example, we might have a chain like:

```ts
[calculateDependenciesPlugin, buildBundlePlugin, finalizeManifestPlugin]
```

where `buildPlugin` is dependent on the context returned by `calculateDependenciesPlugin` and `finalizeManifestPlugin` is dependent on both contextes returned by `calculateDependenciesPlugin` and `buildBundlePlugin`. This makes sense in this strict order, since we hypothetically might  want to write the bundled and css dependencies (from `calculateDependenciesPlugin`) in the manifest, but also use those dependencies to bundle and write the entrypoint (generated by `buildBundlePlugin`) in the manifest.

A plugin probably doesn't want to just try accessing the context for properties it needs, since the chain would throw with an unhelpful message and that would be extremely confusing for whoever is modifying the plugin chain (an infamous example of this is Webpack, where an incorrect configuration of their plugin chain often throws with a generic "cannot access property x of undefined", which is almost always impossible to debug). Some ideas off the top of my head:

1) Validate context in the `validate` step of each plugin and throw with an helpful message if the context doesn't have certain properties. This solution has the disadvantage of being quite verbose and suffers from property clash if the plugins in the chain accidentally overwrite properties.

2) Declare dependencies and let the chain creator factory validate they are respected. This means that whoever develops Modular has to be aware of plugin id / names: changing or creating a new one means manually looking for conflicts and synchronizing existing dependency declarations to align with the new name. IDK how important is it now, but if we want community plugins we also have to be aware of versioning.

What do you guys think?

packages/build-standalone/src/workflow.ts
```diff
@@ -0,0 +1,50 @@
...
+    phase: P,
+    handler: WorkflowHandler<T>,
+  ): this {
+    this.#events.set(phase, [...this.#handlers(phase), handler]);
+    return this;
+  }
+
+  async execute(context: T): Promise<T> {
+    return await series(context, [
+      (c) => series(c, this.#handlers('validate')),
```
cristiano-belloni, line 22: Not clear why `validate` - it seems we're actually doing all the work in the `validate` step in the example plugins.

packages/build-standalone/src/plugins/esbuild-measure-file-size.ts
```diff
@@ -0,0 +1,95 @@
+import recursive from 'recursive-readdir';
+import path from 'path';
+import * as fs from 'fs-extra';
+import { sync as gzipSize } from 'gzip-size';
+
+import type { ModularBuildConfig, ModularBuildContext } from '../types';
+import { canReadAsset, removeFileNameHash } from '../utils';
+import type { WorkflowPlugin } from '../workflow';
+
+export const FILE_SIZES = Symbol('FILE_SIZES');
```
cristiano-belloni, line 10: Why a `Symbol`? The only way of accessing this field for another plugin is to depend on this file and this seems intentional - shouldplugins be coupled like this?

packages/build-standalone/src/workflow.ts
steveukx: This is the definition of the order of the phases referenced in the plugins... it could be rewritten to:

```typescript
const phases: WorkflowStage[] = ['validate', 'cleanup'];

return await series(context, 
  phases.map(phase => (c) => series(c, this.#handlers(phase))
);
```

packages/build-standalone/src/plugins/esbuild-measure-file-size.ts
steveukx: For this particular example, the result of

```
const previousFileSizes = await esbuildMeasureFileSizesBeforeBuild( paths.appBuild );
```

Is only used in:

```
printFileSizesAfterBuild(assets, previousFileSizes);
```

So it makes sense that the data only be accessible to the `esbuildMeasureFileSizes` plugin... note that this plugin hooks into the build lifecycle at two points - the `validate` and `cleanup` phases before and after the build itself.

I would like the plugins to be agnostic of each other, so where possible they would be storing data in the context using a Symbol. Where they really do need to share state, the context should probably have a better way of storing state than the `Record<string, unknown>` type it has at the moment, perhaps by:

- exporting the symbols from the plugins to be used elsewhere
- have a few named properties on the context itself (eg `{ fileSizes?: FileSizes } & Record<string, unknown>`)
- use the plugin factories as keys for the data in a map:

```
function makeWorkflow(config) {
  return createWorkflow(
    doesSomething(config),
    usesSomething(config, doesSomething),
  );
}

function doesSomething(config) {
  return {
    phase: 'validate',
    handler (context) {
      context.get(doesSomething, { /* data */ });
    }
  };
}

function usesSomething(config, doesSomething) {
  return {
    phase: 'cleanup',
    handler (context) {
      const something = context.get(doesSomething);
    }
  };
}
```

PR #2202, opened 2022-11-09
Selective builds: changed packages

cristiano-belloni:
Implement `--changed` and `--compareBranch` for `modular build`

### DIFF
--- a/packages/modular-scripts/src/program.ts
+++ b/packages/modular-scripts/src/program.ts
@@ -84,22 +84,42 @@ program
   )
   .option('--verbose', 'Run yarn commands with --verbose set')
   .option('--private', 'Enable the building of private packages', false)
+  .option(
+    '--changed',
+    'Build only for workspaces that have changed compared to the branch specified in --compareBranch',
+    false,
+  )
+  .option(
+    '--compareBranch <branch>',
+    "Specifies the branch to use with the --changed flag. If not specified, Modular will use the repo's default branch",
+  )
   .action(
     async (
       packagePaths: string[],
       options: {
         preserveModules: string;
         private: boolean;
+        changed: boolean;
+        compareBranch?: string;
       },
     ) => {
       const { default: build } = await import('./build');
       logger.log('building packages at:', packagePaths.join(', '));
 
-      await build(
+      if (options.compareBranch && !options.changed) {
+        process.stderr.write(
+          "Option --compareBranch doesn't make sense without option --changed\n",
+        );
+        process.exit(1);
+      }
+
+      await build({
         packagePaths,
-        JSON.parse(options.preserveModules) as boolean,
-        options.private,
-      );
+        preserveModules: JSON.parse(options.preserveModules) as boolean,
+        private: options.private,
+        changed: options.changed,
+        compareBranch: options.compareBranch,
+      });
     },
   );
 
--- a/packages/modular-scripts/src/build/index.ts
+++ b/packages/modular-scripts/src/build/index.ts
@@ -267,11 +267,26 @@ async function buildStandalone(
   );
 }
 
-async function build(
-  targets: string[],
+async function build({
+  packagePaths: targets,
   preserveModules = true,
-  includePrivate = false,
-): Promise<void> {
+  private: includePrivate,
+  changed,
+  compareBranch,
+}: {
+  packagePaths: string[];
+  preserveModules: boolean;
+  private: boolean;
+  changed: boolean;
+  compareBranch: string;
+}): Promise<void> {
+  if (changed) {
+    console.log(
+      'TODO: implement selective build with changed and compareBranch',
+      changed,
+      compareBranch,
+    );
+  }
   for (const target of targets) {
     try {
       const targetDirectory = await getLocation(target);


ayseenver:
Are the packages a/b/c/d for testing? Do you want to commit them into main?

packages/modular-scripts/src/program.ts
```diff
@@ -84,22 +84,64 @@ program
...
   .option('--verbose', 'Run yarn commands with --verbose set')
   .option('--private', 'Enable the building of private packages', false)
+  .option(
+    '--changed',
+    'Build only for workspaces that have changed compared to the branch specified in --compareBranch',
+    false,
+  )
+  .option(
+    '--descendants',
+    'Additionally build workspaces that descend from the specified packages (can be combined with --changed)',
```
AlbertoBrusa, line 94: Given that the command is called descendants i wouldn't use descend in the description - would be more useful to say 'workspaces which depend on the specified packages' 
Same to --ancestors:
'workspaces that the specified packages depend on'

packages/modular-scripts/src/test/utils.ts
```diff
@@ -132,3 +132,34 @@ export function mockInstallTemplate(
...
   );
 }
+
+/**
+ * Run the Modular cli present at the root path `modularFolder` with a working directory of `cwd`
+ * passing an array of `modularArguments`
+ *
+ * Useful to run modular from source in a different directory (usually a temp directory created for tests)
+ *
+ * @param  modularFolder The root folder in which the modular repo resides
```
AlbertoBrusa, line 142: I feel like the target/cwd could also be a modular repo so I'd make this clearer: 
`@param modularToRun The root modular repo folder containing the modular-scripts to run`

packages/modular-scripts/src/utils/getChangedWorkspaces.ts
```diff
@@ -2,8 +2,22 @@ import path from 'path';
 import pkgUp from 'pkg-up';
 import { getDiffedFiles } from './gitActions';
 import getModularRoot from './getModularRoot';
+import { getAllWorkspaces } from './getAllWorkspaces';
 import type { WorkspaceContent } from '@modular-scripts/modular-types';
 
+// This function returns a WorkspaceContent containing all the changed workspaces, compared to targetBranch
```
AlbertoBrusa, line 8: Should this be in ts-docs format

AlbertoBrusa (commented):
Left a few comments just around general clarity/comments/docs, the code looks good :)

docs/commands/build.md
```diff
@@ -20,6 +21,92 @@ this-is-param-case) in `dist/`
...
+`--descendants`: Build the packages specified by the `[packages...]` argument
+and/or the `--changed` option and additionally build all their descendants (i.e.
+the packages that have a direct or indirect dependency on them) in dependency
+order.
+
+`--ancestors`: Build the packages specified by the `[packages...]` argument
+and/or the `--changed` option and additionally build all their ancestors (i.e.
+the packages that have a direct or indirect dependency on them) in dependency
+order.
+
```
sgb-io, line 46: In the descriptions for `--descendants` and `--ancestors`, what about the scenario where the user doesn't supply any `[packages...]` and doesn't use the `--changed` option?

sgb-io: My assumption is that neither option will have any affect, because all packages are built. Assuming this is correct, might be useful to clarify that in these docs

docs/commands/build.md
```diff
@@ -20,6 +21,92 @@ this-is-param-case) in `dist/`
...
+## Dependency selection and build order examples
+
+We'll be using this configuration of internal (workspace) dependencies in the
+following examples.
+
+```js
+{
```
sgb-io, line 53: Is this example a bit inaccessible to others?

Maybe we should just say `dependencies` instead of `workspaceDependencies`, and throw in a small handul of other things like `react` for example, even though they're not relevant, just to make it a bit closer to a real example.

docs/commands/build.md
cristiano-belloni: `modular build` standalone (unlike `modular test`) will just error because no package is specified. Not sure if it needs to be documented.

packages/modular-scripts/src/program.ts

packages/modular-scripts/src/utils/getChangedWorkspaces.ts
cristiano-belloni: Done

docs/commands/build.md
cristiano-belloni: I added extracts of all the `package.json`s in our imaginary monorepo and mentioned that they were resolved internally to the data structure above (which is easier to read).

sgb-io (approved):

packages/modular-scripts/src/program.ts
```diff
@@ -84,22 +84,64 @@ program
...
+    false,
+  )
+  .option(
+    '--descendants',
+    'Additionally build workspaces that the specified packages directly or indirectly depend on (can be combined with --changed)',
+    false,
+  )
+  .option(
+    '--ancestors',
+    'Additionally build workspaces that are directly or indirectly depended on by the specified packages (can be combined with --changed)',
```
AlbertoBrusa, line 99: Bit nitpicky but:
"Additionally build workspaces that directly or indirectly depend on the specified packages (can be combined with --changed)"
would be clearer - 'depended on by'  had me confused

cristiano-belloni: Yeah, you're right, what I wrote was incorrect. Fixed.

AlbertoBrusa (approved):

sgb-io (approved):

PR #2208, opened 2022-11-15
CMRA: use modular lint instead of eslint directly

sgb-io:
Use `modular lint` instead of a direct `eslint` call for the `lint` script in the create-modular-react-app template

### DIFF
--- a/.changeset/quick-eggs-bow.md
+++ b/.changeset/quick-eggs-bow.md
@@ -1,0 +1,5 @@
+---
+'create-modular-react-app': patch
+---
+
+Use modular lint for linting by default
--- a/packages/create-modular-react-app/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/create-modular-react-app/src/__tests__/__snapshots__/index.test.ts.snap
@@ -72,7 +72,7 @@ exports[`create-modular-react-app WHEN it sets up a project with prefer Offline 
           "private": true,
           "scripts": Object {
             "build": "modular build",
-            "lint": "eslint . --ext .js,.ts,.tsx",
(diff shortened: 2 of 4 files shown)


benpryke (dismissed):

cristiano-belloni:
This needs to fallback to `lint --all` when there is no git repo.

cristiano-belloni (changes requested):
This needs to fallback to `lint --all` when there is no git repo.

sgb-io:
@cristiano-belloni how do you see this working?

To me, it seems that this should be applied in the modular-scripts `lint` command. As you describe, it should detect when there is no git repo, output a warning that the `--all` / `--staged` flags are being overridden and apply `--all`.

One alternative way is to conditionally call `lint` using the appropriate flags from the outside, but this seems fiddly and not consistent with the existing interface of `modular-scripts`. Also, IMO, top-level modular-scripts commands should not throw due to the lack of a git repo.

cristiano-belloni:
> This needs to fallback to `lint --all` when there is no git repo.

Actually: this should be part of selective lint. We can default to all when no argument is supplied and require users to specify `--changed` and (optionally) `--compareBranch` when they want to only lint packages that have changed. There's a possibility this is breaking and needs to go in Modular 5, or we can try to retain backwards compatibility. In any case, linting all packages by default shouldn't be considered a breaking change, because it provides the same functionality (all changed packages are still linted) possibly at the expense of performance (all packages, included the changed ones, are now linted).

cristiano-belloni:
> @cristiano-belloni how do you see this working?

See https://github.com/jpmorganchase/modular/pull/2208#issuecomment-1455903328

PR #2210, opened 2022-11-17
Node 18 & Jest 29

AlbertoBrusa:


package.json
```diff
@@ -5,7 +5,7 @@
     "packages/**"
   ],
   "engines": {
-    "node": "^14.17.0 || >=16.0.0"
+    "node": "^14.18.0 || >=16.10.0 || >=18.0.0"
```
AlbertoBrusa, line 8: Matched node support to what Jest 29 supports

package.json
```diff
@@ -14,7 +14,7 @@
     "lint:fix": "yarn lint --fix",
     "create-modular-react-app": "ts-node packages/create-modular-react-app/src/cli.ts",
     "modular": "ts-node packages/modular-scripts/src/cli.ts",
-    "test": "yarn modular test --watchAll false --runInBand",
+    "test": "yarn modular test --watchAll false --runInBand --env node",
```
AlbertoBrusa, line 17: Our tests are mostly back end tests that don't require the JSDOM env that we set as default for Modular (Modular  users will continue to use JSDOM as default) 
Modular tests that build/run front end components have been changed to specify JSDOM as the env (ignoring the flag passed here)

package.json
```diff
@@ -59,7 +59,7 @@
     "@types/global-modules": "^2.0.0",
     "@types/html-minifier-terser": "6.1.0",
     "@types/is-ci": "3.0.0",
-    "@types/jest": "26.0.24",
+    "@types/jest": "^29.0.0",
```
AlbertoBrusa, line 62: Setting Jest version range to allow minor updates so that we stay up to date until we come across any issues

packages/modular-scripts/src/lint.ts
```diff
@@ -54,11 +55,21 @@ async function lint(
     testPathIgnorePatterns: ['/node_modules/', '/dist/'],
   };
 
-  const testArgs = [
-    ...regexes,
-    '--config',
-    `"${JSON.stringify(jestEslintConfig)}"`,
```
AlbertoBrusa, line 60: We previously passed the Jest config via command line with "" around the object, to make it work on windows 
Since updating Jest, it no longer accepts an object with ' " or ` around it. Removing the quotation marks fixed it for Jest on Mac but breaks it for Windows, as it won't accept it as a valid command.

Actually writing the config file and passing the path works, so I've done that as a workaround, generating it in a temp directory, only if running on Windows.

packages/modular-scripts/src/program.ts
```diff
@@ -157,7 +147,7 @@ program
   .option('--updateSnapshot, -u', testOptions.updateSnapshot.description)
   .option('--verbose', testOptions.verbose.description)
   .option('--watch', testOptions.watch.description)
-  .option('--watchAll [value]', testOptions.watchAll.description, !isCI)
+  .option('--watchAll [value]', testOptions.watchAll.description, false)
```
AlbertoBrusa, line 150: Setting watchAll to true for tests if not running in CI resulted in unexpected Jest behaviour for our users where jest wouldn't exit automatically after a run, so taking the opportunity to change this in modular 4

docs/releases/4.0.x.md
```diff
@@ -0,0 +1,81 @@
...
+
+- [Jest 26 to 27 Change Log](https://github.com/facebook/jest/releases/tag/v27.0.0)
+- [Jest 27 to 28 Migration Guide](https://jestjs.io/docs/28.x/upgrading-to-jest28)
+  | [Change log](https://github.com/facebook/jest/releases/tag/v28.0.0)
+- [Jest 28 to 29 Migration Guide](https://jestjs.io/docs/next/upgrading-to-jest29)
+  | [Change log](https://github.com/facebook/jest/releases/tag/v29.0.0)
+
+Some noteworthy breaking changes:
+
+- Change to default snapshot options to {escapeString: false,
```
cristiano-belloni, line 30: escape the options: `{escapeString: false, printBasicPrototype: false}`

packages/modular-scripts/src/esbuild-scripts/plugins/svgr.ts
```diff
@@ -8,7 +8,23 @@ import getModularRoot from '../../utils/getModularRoot';
 import { normalizeToPosix } from '../utils/formatPath';
 
 const svgrOptions: svgr.Config = {
-  template(variables, { tpl }) {
+  template(
```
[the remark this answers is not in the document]
cristiano-belloni: I guess related to https://github.com/jpmorganchase/modular/pull/1193 (for future documentation)

packages/modular-scripts/src/test/index.ts
```diff
@@ -85,14 +87,28 @@ async function test(
...
+  // pass in jest configuration
   const { createJestConfig } = await import('./config');
-  cleanArgv.push(
-    '--config',
-    `"${JSON.stringify(
+
+  // Only used when running on Windows
+  const tempConfigPath = path.join(tmp.dirSync().name, 'temp-jest-config.json');
+
+  // Windows: pass in configuration as a file as jest won't accept "" around the object and Windows doesn't accept JSON object {} passed directly
```
cristiano-belloni, line 96: We now have two copies of this code with the Windows temp hack; can we factor it out in a `invokeJest` function?

AlbertoBrusa: Yep good point, will do

cristiano-belloni (approved):

packages/modular-scripts/src/test/index.ts
AlbertoBrusa: Done

cristiano-belloni:
LGTM; please at some point remember to merge `main` back in (it can be in this branch or in `feature/v4`, but better soon than later)

PR #2214, opened 2022-11-24
[WIP] feat: add code splitting to esbuild config

ayseenver:


cristiano-belloni (changes requested):
I took a look and it seems our `format` is hardcoded to `esm`, so this setting should work.

This is the work that I think we need to do:

1. Move this functionality behind a flag
2. Write one or more test cases in `app.test.ts` where we build with esbuild an app whose source code does a dynamic import of some other file in the project and see if the flag:
  - effectively builds the app into more than one chunks
  - allows the app to get `start`ed correctly by Modular in esbuild mode
3. We also need to test how this works with `esm-view`s to decide whether to restrict this option to only `app`s, but that can be done manually after completing the previous points

PR #2216, opened 2022-11-25
Update to ESLint 8 & Supported TypeScript to >4.5.3

AlbertoBrusa:
Jest 29 JSDOM testing env requires TypeScript 4.5 or greater, ESLint 8 adds support for TypeScript 4.5 or greater.
To support Jest 29, we're upgrading ESLint to 8 and finally freeing the TypeScript version constraints (previously locked)

In the updating dependencies process I also re-generated our lock file hence the updated browser target snapshots


packages/modular-scripts/src/esbuild-scripts/start/runtime/index.ts
```diff
@@ -59,7 +59,7 @@ interface WebSocketMessage {
 }
 
 connection.onmessage = (m: MessageEvent) => {
-  const message = JSON.parse(m.data) as WebSocketMessage;
```
sgb-io, line 62: Might be worth having a runtime safety check here. The simplest upgrade is to make it throw when `m.data` isn't a string - since `JSON.parse` will already throw in that situation. 

I'm not sure what the consequences are of `m.data` being non-string - those consequences should drive if this should throw or warn (or do nothing, or something else).

I say all this on the assumption that the `MessageEvent` interface is accurate - if `data` can be non-string, our code should account for that, probably (or we should known why we *don't* need to account for it).

packages/modular-scripts/src/__tests__/start-app.ts
```diff
@@ -113,6 +113,7 @@ export async function startApp(
   // starts above, but don't want to wait until the
   // process itself finishes.
   return {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
cristiano-belloni, line 116: Why is this promise "misused" and not "floating"? Asking because I think we can silence the "floating" error without eslint comments, by just prepending `void`.

packages/modular-scripts/src/utils/getPackageDependencies.ts
```diff
@@ -82,6 +82,7 @@ export async function getPackageDependencies(target: string): Promise<{
 
   // Package dependencies can be either local to the package or in the root package (hoisted)
   const packageDeps = Object.assign(
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
cristiano-belloni, line 85: Is this because `Object.create(null)` is typed as `any`?

packages/modular-scripts/src/utils/memoize.ts
```diff
@@ -5,6 +5,7 @@ export default function memoize<R, T extends (...args: any[]) => R>(f: T): T {
 
   const g = (...args: any[]) => {
     if (!memory.has(args.join())) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
cristiano-belloni, line 8: I think you can avoid this by using `unknown` instead of `any`?

packages/modular-site/src/App.tsx
```diff
@@ -20,6 +20,7 @@ import './App.css';
 const worker = perspective.shared_worker();
 
 async function getTable(): Promise<Table> {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
cristiano-belloni, line 23: Not sure we should even *have* modular-site tbh :)

AlbertoBrusa: Shall I leave it like this for now as we're removing it?

packages/modular-scripts/src/utils/getPackageDependencies.ts
AlbertoBrusa: Yep so: ``Unsafe argument of type `any` assigned to a parameter of type `object`.``
Object.create() returns any 

what would be the cleanest way to deal with this. Using type assertion, typescript suggests:

_Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`._
- _If you want a type meaning "any object", you probably want `Record<string, unknown>` instead._
- _If you want a type meaning "any value", you probably want `unknown` instead._

so `Object.create(null) as Record<string, unknown>,` works but doesn't look very clean

cristiano-belloni: This is unfortunately a Typescript "needs proposal" that is currently open: https://github.com/microsoft/TypeScript/issues/1108

People had luck with typing is as `void` or `null`, it seems.

packages/modular-site/src/App.tsx
cristiano-belloni: 👍

packages/modular-scripts/src/program.ts
```diff
@@ -235,7 +235,7 @@ program
     const { default: initWorkspace } = await import('./init');
```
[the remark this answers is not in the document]
AlbertoBrusa: I was curious on why we weren't taking a boolean in the first place and as I suspected setting the type of preferOffline to boolean means that`--prefer-offline false` still works as intended, without having to check ourselves that a boolean was provided

packages/modular-scripts/src/utils/getPackageDependencies.ts
cristiano-belloni: Or, more correctly:

```ts
const nullObject: Record<keyof Object, never> = Object.create(null)
```

packages/modular-scripts/src/esbuild-scripts/start/runtime/index.ts
AlbertoBrusa: Thanks, I tried to improve my understanding of .data and why it's typed as any/if we need to consider it being anything but string. 

I opted to follow your suggestion to throw if not string, as JSON.parse will have done anyways. hopefully should never be the case.

sgb-io (approved):

cristiano-belloni (approved):

PR #2218, opened 2022-11-28
Fix repository field generation on Windows

joshwooding:


packages/modular-scripts/src/build/buildPackage/getRepositoryField.ts
```diff
@@ -11,7 +11,7 @@ async function getRelativePathInRepo(packagePath: string) {
     ['rev-parse', '--show-toplevel'],
     { stdout: 'pipe' },
   );
-  return path.relative(gitRepoPath, packagePath);
+  return path.relative(gitRepoPath, packagePath).replace(/\\/g, '/');
```
cristiano-belloni, line 14: Can you please use `path.win32.sep` and `path.posix.sep`?

cristiano-belloni (changes requested):
It seems node 14 is failing with `path.relative(...).replaceAll is not a function`.
This is because we transpile, but not polyfill.

packages/modular-scripts/src/build/buildPackage/getRepositoryField.ts
```diff
@@ -11,7 +11,10 @@ async function getRelativePathInRepo(packagePath: string) {
     ['rev-parse', '--show-toplevel'],
     { stdout: 'pipe' },
   );
-  return path.relative(gitRepoPath, packagePath);
+  return path
+    .relative(gitRepoPath, packagePath)
+    .split(path.win32.sep)
```
joshwooding, line 16: This is the same as the utils in esbuild-scripts

joshwooding:
Any update on this?

cristiano-belloni (approved):

PR #2227, opened 2022-12-06
Use Config file instead of Environment Variables 

AlbertoBrusa:
Move modular configuration away from env variables and into a configuration file

- [x] Read .modular.js config in modular project root to load up configuration
- [x] Allow env variables to override configuration fill
- [x] Allow use of config where ENV Variables are currently used:
  - [x] USE_MODULAR_ESBUILD
  - [x] ~~USE_MODULAR_WEBPACK~~ 
  - [x] Remove USE_MODULAR_WEBPACK
  - [x] EXTERNAL_CDN_TEMPLATE
  - [x] EXTERNAL_ALLOW_LIST
  - [x] EXTERNAL_BLOCK_LIST
  - [x] PUBLIC_URL
  - [x] GENERATE_SOURCEMAP
      - [x] Even where we don't currently use the ENV 
- [x] Test: 
   - [x] Reading config file
   - [x] ENV variable override
- [x] Change default CDN to esm.sh 
- [x] Document configuring via configuration file
   - [x] Document USE_MODULAR_ESBUILD only affects view/esm-view
   - [x] CDN_TEMPLATE only affects esm views
   - [x] Discourage allow/deny lists 
- [x] Change all references to env variables in documentation to mention config file
- [x] Edit 4.0.0 changes documentation

We have some of the ENV variables read in .js files that can't use our TS config logic for some reason, so we'll set an ENV variable prefixed with INTERNAL_ for those files to read

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -7,6 +7,81 @@ nav_order: 10
 Modular has minimal configuration because of its philosophy. However there is a
 set of minimum configuration required.
 
+## Configuration File
+
+We allow a number of Modular behaviours to be configured via a dedicated Modular
+config file, `.modular.js`, located at the root of the repository.
+
+We support the following file names/formats:
+
+- `modular` property within your `package.json`
+- `.modular.js`
+- `.modularrc`
+- `.modularrc.json`
+- `.modularrc.yaml`
+- `.modularrc.yml`
+- `.modularrc.js`
+- `.modularrc.cjs`
+- `modular.config.js`
+- `modular.config.cjs`
+
+Example `.modular.js` file contents with all configurable attributes and their
+default values:
+
+```js
+module.exports = {
+  useModularEsbuild: false,
+  externalCdnTemplate: 'https://esm.sh/[name]@[version]',
+  externalBlockList: [],
+  externalAllowList: ['**'],
+  publicUrl: '',
+  generateSourceMap: true,
+};
+```
+
+### useModularEsbuild
+
+`boolean`
+
+Use esbuild instead of default Webpack. Only affects Views and ESM Views.
+
+### externalCdnTemplate
+
+`string`
+
+Template to resolve the URL used to fetch packages from a CDN. Defaults to
+esm.sh. Only applies to ESM Views.
+
+### externalBlockList
+
+`string[]`
+
+Packages that should be bundled and not fetched from a CDN. Avoid using this
+unless absolutely necessary. Defaults to none. Only applies to ESM Views.
+
+### externalAllowList
+
+`string[]`
+
+Packages that should be fetched from a CDN. Avoid changing this unless
+absolutely necessary. Defaults to all packages. Only applies to ESM Views.
+
+### publicUrl
+
+`string`
+
+Same as Create React App PUBLIC_URL. Instead of assuming the application is
+hosted in the web server's root or subpath specified by homepage in
+package.json, assets will be referenced to the URL provided.
+
+### generateSourceMap
+
+`boolean`
+
+Should build process generate a source map - can be disabled for performance
+reasons. Source maps are resource heavy and can cause out of memory issue for
+large source files.
+
 ## `package.json#modular`
 
 _NOTE: This property is created automatically and is described here for
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -39,44 +39,56 @@ module.exports = {
 };
 ```
 
-### useModularEsbuild
+### **useModularEsbuild**
 
-`boolean`
+**Type**: `boolean`
+
+**Default**: `false` - Uses [Webpack](https://webpack.js.org/)
 
 Use esbuild instead of default Webpack. Only affects Views and ESM Views.
 
-### externalCdnTemplate
+### **externalCdnTemplate**
 
-`string`
+**Type**: `string`
+
+**Default**: `https://esm.sh/[name]@[version]` - [esm.sh](https://esm.sh/)
 
 Template to resolve the URL used to fetch packages from a CDN. Defaults to
 esm.sh. Only applies to ESM Views.
 
-### externalBlockList
+### **externalBlockList**
 
-`string[]`
+**Type**: `string[]`
+
+**Default**: `[]` - No packages
 
 Packages that should be bundled and not fetched from a CDN. Avoid using this
 unless absolutely necessary. Defaults to none. Only applies to ESM Views.
 
-### externalAllowList
+### **externalAllowList**
 
-`string[]`
+**Type**: `string[]`
+
+**Default**: `[**]` - All packages
 
 Packages that should be fetched from a CDN. Avoid changing this unless
 absolutely necessary. Defaults to all packages. Only applies to ESM Views.
 
-### publicUrl
+### **publicUrl**
 
-`string`
+**Type**: `string`
+
+**Default**: `''` - No Public URL
 
 Same as Create React App PUBLIC_URL. Instead of assuming the application is
 hosted in the web server's root or subpath specified by homepage in
 package.json, assets will be referenced to the URL provided.
 
-### generateSourceMap
+### **generateSourceMap**
 
-`boolean`
+**Type**: `boolean`
+
+**Default**: `true`
 
 Should build process generate a source map - can be disabled for performance
 reasons. Source maps are resource heavy and can cause out of memory issue for
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,8 +4,9 @@ nav_order: 10
 
 # Configuration
 
-Modular has minimal configuration because of its philosophy. However there is a
-set of minimum configuration required.
+Because of its philosophy, Modular has a restricted set of configurable
+behaviours. Additionally it requires some minimal configuration within the
+`package.json`s in the repository, all handled by Modular itself.
 
 ## Configuration File
 
@@ -62,8 +63,11 @@ esm.sh. Only applies to ESM Views.
 
 **Default**: `[]` - No packages
 
-Packages that should be bundled and not fetched from a CDN. Avoid using this
-unless absolutely necessary. Defaults to none. Only applies to ESM Views.
+Packages that should be bundled and not fetched from a CDN. We recommend
+allowing all packages to be handled by the CDN, except for particular cases
+where they would not work correctly. See
+[known-limitations](./esm-views/known-limitations.md). Defaults to none. Only
+applies to ESM Views.
 
 ### **externalAllowList**
 
@@ -71,8 +75,10 @@ unless absolutely necessary. Defaults to none. Only applies to ESM Views.
 
 **Default**: `[**]` - All packages
 
-Packages that should be fetched from a CDN. Avoid changing this unless
-absolutely necessary. Defaults to all packages. Only applies to ESM Views.
+Packages that should be fetched from a CDN. We recommend allowing all packages
+to be handled by the CDN, except for particular cases where they would not work
+correctly. See [known-limitations](./esm-views/known-limitations.md). Defaults
+to all packages. Only applies to ESM Views.
 
 ### **publicUrl**
 
--- a/docs/esm-views/customize-bundle-strategy.md
+++ b/docs/esm-views/customize-bundle-strategy.md
@@ -10,13 +10,14 @@ By default, all external dependencies are rewritten to a CDN URL and none is
 bundled. This logic can be controlled using two environment variables or by
 using a [modular configuration file](../configuration.md).:
 
-1. `EXTERNAL_ALLOW_LIST` is a comma-separated string that specifies which
-   dependencies are allowed to be rewritten to the CDN; if not specified, its
-   default value is `**` ( -> all dependencies are rewritten)
-2. `EXTERNAL_BLOCK_LIST` is a comma-separated string that specifies which
-   dependencies are **not** allowed to be rewritten to the CDN; if not specified
-   its default value is empty ( -> no dependency excluded, i.e. all dependencies
-   are rewritten)
+1. [`EXTERNAL_ALLOW_LIST`](../configuration.md#externalallowlist) is a
+   comma-separated string that specifies which dependencies are allowed to be
+   rewritten to the CDN; if not specified, its default value is `**` ( -> all
+   dependencies are rewritten)
+2. [`EXTERNAL_BLOCK_LIST`](../configuration.md#externalblocklist) is a
+   comma-separated string that specifies which dependencies are **not** allowed
+   to be rewritten to the CDN; if not specified its default value is empty ( ->
+   no dependency excluded, i.e. all dependencies are rewritten)
 
 The allow / block lists are parsed and processed according to this logic:
 
--- a/docs/esm-views/customize-bundle-strategy.md
+++ b/docs/esm-views/customize-bundle-strategy.md
@@ -7,7 +7,8 @@ title: Customize bundling strategy
 # Customize bundling / rewriting strategy
 
 By default, all external dependencies are rewritten to a CDN URL and none is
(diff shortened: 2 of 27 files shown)


packages/modular-scripts/src/utils/config.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+/**
+ * Reads env variables and configuration to understand if esbuild or Webpack should be used
+ * Webpack is used as default if no configuration is set or both are set to true
+ * Environment variables take precedence over config file
+ * @returns True if esbuild should be used or false if webpack should be used
+ */
+export async function utilizeEsbuild(): Promise<boolean> {
```
cristiano-belloni, line 67: Do we need a function per configuration switch? I'd be more inclined to have a config object like this:

```ts
const config = {
  useModularEsbuild: {
    default: false,
    override: process.env.USE_MODULAR_ESBUILD === 'true'
  } 
}
```

and a function that is called like `getConfig("useModularEsbuild")`.

This means that you need to do `getConfig("useModularEsbuild") || getConfig("useModularWebpack")` in the code, but TBH I'm happy to remove `USE_MODULAR_WEBPACK` (which nobody uses it as it's the default + is not documented anyway).

packages/modular-scripts/src/utils/config.ts
```diff
@@ -0,0 +1,91 @@
...
+  const loadedConfiguration = await configuration;
+  // Handle no or empty configuration - debug log? don't think we should error should we?
+  if (loadedConfiguration) {
+    // Error if configuration doesn't match our interface?
+    const config = loadedConfiguration.config as ConfigObject;
+    const value = config[configEntry];
+    if (value) {
+      return value;
+    } else {
+      throw new Error(
```
cristiano-belloni, line 50: I think users can always omit config entries (that's why we need defaults, see comment below)

packages/modular-scripts/src/utils/config.ts
AlbertoBrusa: Thanks for the config object approach, rewrote around it and it's much better now

cristiano-belloni:
@AlbertoBrusa - While we're at it, can we set `EXTERNAL_CDN_TEMPLATE`'s default to `'https://esm.sh/[name]@[resolution]'`

packages/modular-scripts/react-scripts/config/getHttpsConfig.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
+require('ts-node').register();
```
cristiano-belloni, line 3: I think @steveukx was referring to running the project with `node -r ts-node/register`, not using it programatically. I run it with `ts-node` directly, which does the smae preloading and should allow you to import ts files in your js files:

`alias modular-dev="~/modular/node_modules/.bin/ts-node ~/modular/packages/modular-scripts/src/cli.ts"`

There is an important caveat, though: if you try to require ts files in a js file that is loaded by Webpack (for example, anything required by webpack.config.js), this won't work, because it will use a different instance of node spawned directly by Webpack. In this case, `getHttpsConfig` is required by `start`, so it should work.

(for Webpack files, you can use `TS_NODE_PROJECT`, as @steveukx just discovered).

AlbertoBrusa: Ah okay - I thought all our scripts already ran with ts-node, for example tests: `ts-node packages/modular-scripts/src/cli.ts test --watchAll false --runInBand --env node` - and that isn't happy with importing a ts file in js, while when running it programmatically it didn't complain (tho I'm still facing other problems)

We can chat more about it, and see what to do for Webpack as well

AlbertoBrusa: Ignore all of the above I went down the wrong rabbit hole.
I'm still unclear on what the 'right' way of importing a typescript function in a javascript file is

cristiano-belloni: @AlbertoBrusa just for fun, you can also try to use tsx (which in turn uses esbuild transform) instead of ts-node. I don't know if it solves your problem, but worth checking: https://github.com/esbuild-kit/tsx

cristiano-belloni: @AlbertoBrusa you can also try ts-node in ESM mode -> https://github.com/TypeStrong/ts-node#esm (if you haven't already)

cristiano-belloni: Finally, people from the internet say that ignoring warnings might work -> `node --no-warnings --loader ts-node/esm file.ts`

AlbertoBrusa: Thanks, I'll give all these approaches a try tomorrow

packages/modular-scripts/package.json
```diff
@@ -36,6 +36,7 @@
     "@rollup/plugin-node-resolve": "13.3.0",
     "@svgr/core": "6.2.1",
     "@svgr/webpack": "6.2.1",
+    "@types/lodash": "^4.14.191",
```
AlbertoBrusa, line 39: Not sure in the slightest why that popped up now but it was complaining for missing typings for lodash

packages/modular-scripts/react-scripts/config/paths.js
```diff
@@ -46,7 +46,9 @@ const resolveModular = (relativePath) =>
 const publicUrlOrPath = getPublicUrlOrPath(
   process.env.NODE_ENV === 'development',
   require(resolveApp('package.json')).homepage,
-  process.env.PUBLIC_URL,
+  process.env.INTERNAL_PUBLIC_URL === ''
```
AlbertoBrusa, line 49: letting getConfig return undefined would have been messy, but treating '' (empty string) as configuration not set might not be ideal either, open to suggestions

packages/modular-scripts/src/build/buildPackage/makeBundle.ts
```diff
@@ -107,7 +108,7 @@ export async function makeBundle(
 
   const outputOptions: rollup.OutputOptions = {
     freeze: false,
-    sourcemap: true, // TODO: read this off env
+    sourcemap: getConfig('generateSourceMap') as boolean,
```
AlbertoBrusa, line 111: The `as boolean` is safe, as we know for sure what getConfig for each configuration field will return

docs/configuration.md
```diff
@@ -7,6 +7,81 @@ nav_order: 10
...
+
+`string`
+
+Same as Create React App PUBLIC_URL. Instead of assuming the application is
+hosted in the web server's root or subpath specified by homepage in
+package.json, assets will be referenced to the URL provided.
+
+### generateSourceMap
+
+`boolean`
```
steveukx, line 79: Would it be possible to include details of the default values here?

docs/esm-views/customize-bundle-strategy.md
```diff
@@ -7,7 +7,8 @@ title: Customize bundling strategy
 # Customize bundling / rewriting strategy
 
 By default, all external dependencies are rewritten to a CDN URL and none is
-bundled. This logic can be controlled using two environment variables:
+bundled. This logic can be controlled using two environment variables or by
+using a [modular configuration file](../configuration.md).:
 
 1. `EXTERNAL_ALLOW_LIST` is a comma-separated string that specifies which
```
steveukx, line 13: You could link to the specific option here...

```suggestion
1. [EXTERNAL_ALLOW_LIST](../configuration.md#externalallowlist) is a comma-separated string that specifies which
```

AlbertoBrusa: Good suggestion

docs/releases/4.0.x.md
```diff
@@ -66,6 +67,13 @@ manually complete the tasks previously covered by these commands.
 - Dropped support for minor versions of Node 14.17 and Node 16 version 16.9 and
   below
   - Now support Node ^14.18.0, >=16.10.0, and >=18.0.0
+- Dropped `USE_MODULAR_WEBPACK` environment variable, as Webpack is used by
+  default. Use `USE_MODULAR_ESBUILD` env variable or useModularEsbuild in a
```
steveukx, line 71:
```suggestion
  default. Use `USE_MODULAR_ESBUILD` env variable or `useModularEsbuild` in a
```

docs/configuration.md
```diff
@@ -7,6 +7,93 @@ nav_order: 10
 Modular has minimal configuration because of its philosophy. However there is a
```
sgb-io, line 7: We should update the wording here now IMO. It's a bit confusing saying "there is a set of minimum configuration required" and then going on to talk about some optional configuration.

Maybe something like:

```markdown
Modular has minimal configuration because of its philosophy, but some behaviours can be controlled.
```

docs/configuration.md
```diff
@@ -7,6 +7,93 @@ nav_order: 10
...
+Template to resolve the URL used to fetch packages from a CDN. Defaults to
+esm.sh. Only applies to ESM Views.
+
+### **externalBlockList**
+
+**Type**: `string[]`
+
+**Default**: `[]` - No packages
+
+Packages that should be bundled and not fetched from a CDN. Avoid using this
```
sgb-io, line 65: I think "Avoid using this unless absolutely necessary" is a bit vague. Can we give some clearer guidance?

Perhaps something like

```markdown
It is not recommended to change the default value, but this option exists to allow special handling of certain dependencies, e.g. `react`.
```

sgb-io (commented):
Will resume my review tomorrow, rather than keeping this open until then

docs/configuration.md
AlbertoBrusa: Good point, I've now tried to add some details and linked the documentation page containing all details regarding CDN Limitations

packages/modular-scripts/src/utils/config.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @returns configured value:
+ * - the override environment variable if configured
+ * - the value stated in the config file if provided
+ * - the default value if neither environment variable nor the config file are provided
+ *
+ * Although return type can be many things, we can use 'as' with confidence to restrict it to the type we're querying:
+ *
+ * - 'useModularEsbuild' will always be a boolean, so we can do getConfig('useModularEsbuild') as boolean
+ */
+export function getConfig(
```
steveukx, line 92: By using generics in this function you can type the return so you don't need to cast the return type in the caller:

```typescript
interface Config {
  a: string | null;
  b: number | null;
}

type ConfigDefs = {
  [Key in keyof Config]: {
    default: Exclude<Config[Key], null>;
    override: undefined | Exclude<Config[Key], null>;
  }
}

const defs: ConfigDefs = {
  a: {
    default: 'a',
    override: undefined,
  },
  b: {
    default: 1,
    override: 2,
  },
};

function getConfig<T extends keyof ConfigDefs>(key: T): Exclude<Config[T], null> {
  return defs[key].override ?? defs[key].default;
}
}
```

packages/modular-scripts/src/utils/config.ts
```diff
@@ -0,0 +1,108 @@
...
+  `.modularrc.js`,
+  `.modularrc.cjs`,
+  `modular.config.js`,
+  `modular.config.cjs`,
+];
+
+// Look for configuration file
+const modularRoot = getModularRoot();
+const explorer = cosmiconfigSync('modular', { searchPlaces });
+const configResult = explorer.search(path.join(modularRoot, 'package.json'));
```
steveukx, line 22: If you were to cast the type here, you can remove the inline type assertions in the `getConfig` below:

```suggestion
const configResult: undefined | { config: Partial<ConfigObject> }  = explorer.search(path.join(modularRoot, 'package.json'));
```

Using a `Partial<T>` in this type would mean you can remove the `| null` from each property of `ConfigObject` along with all the `NonNullable<T>` inline types

packages/modular-scripts/src/utils/config.ts
AlbertoBrusa: Thanks! I was hoping there was a clever way of doing that, looks much cleaner now. Learn something new every day.

packages/modular-scripts/src/utils/config.ts
AlbertoBrusa: I've given it a go, much cleaner now, but doesn't feel quite 'perfect' yet, feels like there should be a better way of doing: 
```ts
  let configValue;
  if (configResult) {
    const loadedConfigValue = configResult.config[key];
    if (typeof loadedConfigValue === typeof defs[key].default) {
      configValue = loadedConfigValue as Exclude<Config[T], null>;
    }
  }
  return defs[key].override ?? configValue ?? defs[key].default;
```

AlbertoBrusa: Updated attempt:
```js
export function getConfig<T extends keyof ConfigDefs>(key: T): Config[T] {
  const configValue: Config[T] | undefined = configResult
    ? configResult.config[key]
    : undefined;
  return defs[key].override ?? configValue ?? defs[key].default;
}
```

quite happy with that

steveukx: I used the two types in my suggestion, but you should probably collapse to just the one type in the `getConfig` method:

```typescript
export function getConfig<T extends keyof Config>(key: T): Config[T] {
```

packages/modular-scripts/src/utils/config.ts
```diff
@@ -0,0 +1,98 @@
...
+        : undefined,
+  },
+  externalCdnTemplate: {
+    default: 'https://esm.sh/[name]@[version]',
+    override: process.env.EXTERNAL_CDN_TEMPLATE,
+  },
+  externalBlockList: {
+    default: [],
+    override: process.env.EXTERNAL_BLOCK_LIST
+      ? process.env.EXTERNAL_BLOCK_LIST.split(',')
```
sgb-io, line 57: Maybe it's not a big deal, but I guess this lacks any validation at the moment, so could throw or lead to unexpected behaviour if someone provides unexpected values. For example something like `foo,,`.

The boolean and single-string cases are a lot simpler and I can't see anything breaking there, it's just the use of "if there is any value for X, assume I can do `split(',')` on it" that might be a touch unsafe

packages/modular-scripts/react-scripts/config/paths.js
sgb-io: Not sure what you mean here. Why is this value not consumed via `getConfig`?

sgb-io (commented):
Everything else LGTM

packages/modular-scripts/src/utils/config.ts
AlbertoBrusa: good shout, I'll see if I can do some light validation. 
This is how it currently behaves in released modular, but no reason we can't improve it and make it safer

sgb-io: If we have the guarantee that the value is a string (empty or not), that helps, but I wonder if ending up with empty items in the output could lead to unexpected behaviour. Maybe just check that every time in the `split()` outcome is a string with .length of >0, otherwise, potentially throw? It could be helpful to tell the user that their config might be wrong

(That said we don't want to throw when it's simply un-set or an empty string!)

AlbertoBrusa: I was adding validation but I think the context of this logic makes it okay without:
Process.env always returns a string, so we're not worried about the type. 
if there's no ',' to split on, we take the whole string as a package name 
If we split on the ',' and we get empty strings, they won't match with any packages, and won't cause any damage
So the worst case scenario is that they enter the env variable incorrectly and don't get the exclusion/inclusion they wanted in their block list - but nothing will break

sgb-io (approved):

PR #2228, opened 2022-12-06
modular `source` type

cristiano-belloni:
## `modular.type: source` and `--dangerouslyIgnoreCircularDependencies`

This PR introduces the `source` type (i.e., a package that can be imported but never gets built or started)

### Why we are considering `source` packages

The `source` type is useful for all those packages that just get imported from `app`, `esm-view` or `package` workspaces  and never get built (ie common utils or views only used internally). When the `package` type is used for these kind of workspaces, a confusing situation arises, in which these packages should never be built standalone (they can even fail, since they're often intended to be built as part of `app`s or `esm-view`s and might need build-time plugins) but still get built by selective build operators (for example, `--changed` will try to incrementally build all changed Modular workspaces). There are two ways of avoiding this: disallowing direct imports from source via eslint rules or introducing the `source` type. The first solution is not really practical, since many pre-existing projects are already set-up to import other packages from source (and you probably shouldn't be forced to build all your common utils everytime), so the `source` type could be an useful solution to the problem.

### When to convert to `source` packages

If your workspace:

1) has type `package`*
2) is imported by more than one `app`, `esm-view` or `package` type in your monorepo**
3) is always imported from its source, not from its `dist` directory
4) doesn't ever need to be built standalone (e.g. it's never published to a package registry or only its code is)

your workspace is a good candidate to be converted to the `source` type

\* You can also convert a `view` to `source`, but you will lose the possibility of `modular start`ing it.
\*\* If it's imported by only one workspace and it's not published as per 4), you might consider incorporating it into the importing package?

## `source` packages and circular dependencies

Normally `modular build` throws when there are circular dependencies in the workspace dependencies graph, since it can't calculate the correct build order.  Since `source` types don't build, however, it's possible for Modular to work with circular dependencies, _if the circularity depends on `source` packages_ (i.e., if, removing source packages from the graph, the cycle[s] disappear). We require the flag `--dangerouslyIgnoreCircularDependencies` to be set in that case.

### Why we are considering allowing circular dependencies

Circular dependencies should not be used; they [mess up the require order](https://stackoverflow.com/questions/10869276/how-to-deal-with-cyclic-dependencies-in-node-js), they can confuse developer tools and they are always fixable [by creating additional dependencies which contain the common parts](https://nx.dev/recipes/other/resolve-circular-dependencies).
When Modular analyses dependencies to build them in order, the algorithm can't decide what to build first when it encounters a circular dependency (e.g.: if A depends on B but B depends on A, how do you know which `dist` files are needed first?), and it throws. Even if A imports `B/a-module` and B depends on `A/some-module` (so the circularity can technically "work", because the code that  imports `a-module` isn't the same code that is imported by `B`, and vice-versa), it's impossible to determine that that is the case by just statically analysing the code. Additionally, the resulting code is a nightmare to maintain (if you refactor files in A so that `import` and `import`-ed are in the same file, the build chain comes crashing down). If possible, projects should not use circular dependencies, and Modular might to want warn in the future when it detects them.

This PR shows that the Modular order algorithm can allow circular dependencies if we introduce a new `modular.type` called `source`, which signals that the package doesn't need to (and cannot) be built. If we have a circular dependency involving a source package (i.e. if `B` is `source` in the previous example), we can still calculate a valid build order by removing `source`s from the order graph, since they don't get built and nothing can depend on their build result anyway. This is not to encourage circular deps, but to allow huge projects that unfortunately already have them to be onboarded without extensive refactoring.

### Cycle disappearing when `source` types are removed from the dependency graph (cycle between `package` and `source`)

(assuming that `package` b is depending on `source` c and `source` c is
depending on `package` b and `package` d):

#### Without `--dangerouslyIgnoreCircularDependencies` the build fails

```bash
> modular-dev build b --descendants
[modular] Building packages at: b
[modular] Cycle detected, b -> c -> b
```

#### With `--dangerouslyIgnoreCircularDependencies` the build warns but succeeds

```bash
> modular-dev build b --descendants --dangerouslyIgnoreCircularDependencies
[modular] Building packages at: b
[modular] You chose to dangerously ignore cycles in the dependency graph. Builds will still fail if a cycle is found involving two or more buildable packages. Please note that the use of this flag is not recommended. It's always possible to break a cyclic dependency by creating an additional dependency that contains the common code.
[modular] $ b: generating .d.ts files
[modular] $ b: building b...
[modular] $ b: built b in /Users/N761472/dev/rig/ghost-building/dist/b
[modular] $ d: generating .d.ts files
[modular] $ d: building d...
[modular] $ d: built d in /Users/N761472/dev/rig/ghost-building/dist/d
```

### Cycle not disappearing when `source` types are removed from the dependency graph (cycle between `package`s)

(assuming that `package` b is depending on `package` c and `package` c is
depending on `package` b and `package` d):

#### Even with `--dangerouslyIgnoreCircularDependencies` the build fails:

```bash
> modular-dev build b --descendants --dangerouslyIgnoreCircularDependencies
[modular] Building packages at: b
[modular] You chose to dangerously ignore cycles in the dependency graph. Builds will still fail if a cycle is found involving two or more buildable packages. Please note that the use of this flag is not recommended. It's always possible to break a cyclic dependency by creating an additional dependency that contains the common code.
[modular] Cycle detected, b -> c -> b
```

## TODO

- [x] implement `source` type
- [x] implement removal from graph
- [x] implement `--dangerouslyIgnoreCircularDependencies`
- [x] document circular dependencies
- [x] document source type
- [x] test `source` type
- [x] test `--dangerouslyIgnoreCircularDependencies`


packages/modular-scripts/src/utils/packageTypes.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface PackageTypeDefinition {
+  buildable: boolean;
+  testable: boolean;
+  startable: boolean;
+}
+
+type PackageTypeDefinitions = { [Type in PackageType]: PackageTypeDefinition };
+
+const packageTypeDefinitions: PackageTypeDefinitions = {
```
sgb-io, line 18: Ideal content to go in the markdown docs as a table

sgb-io (commented):
Looks good so far

packages/modular-scripts/src/utils/selectWorkspaces.ts
```diff
@@ -68,14 +81,46 @@ export async function selectWorkspaces({
...
   }
 
-  packageScope = [...new Set(packageScope)];
+  // We want to remove all the non-buildable packages from the package scope. Create a filter predicate that looks up to the package map
+  const isWorkspaceBuildable = (name: string) => {
+    const type = allWorkspacePackages.get(name)?.modular?.type;
+    return type && isBuildableModularType(type as PackageType);
+  };
+
+  packageScope = [...new Set(packageScope)].filter(isWorkspaceBuildable);
```
cristiano-belloni, line 90: Move this inside the if, or split the function in two. We need to select non-buildable packages in case we use this to select testable packages.

docs/concepts/circular-dependencies.md
```diff
@@ -0,0 +1,78 @@
...
+of B depends on an unrelated part of A, refactoring code in A can make the
+require order in B fail and vice versa.
+
+## `--dangerouslyIgnoreCircularDependencies` escape hatch
+
+If your circular dependencies involve packages that never get built (namely,
+`modular.type: source` packages), Modular can still calculate the correct build
+order by removing them from the dependency graph. By default, `modular build`
+will still refuse to build any build graph that contains a circular dependency,
+but this behaviour can be overrided by specifying the
```
AlbertoBrusa, line 30: *overridden

AlbertoBrusa (commented):
Looks good to me

docs/concepts/circular-dependencies.md
cristiano-belloni: Thanks, done!

AlbertoBrusa:
Approved, but wondering if we should PR into feature/v4 and fixing the conflicts (if any) now- having said that I don't mind merging to main and then pulling main into v4 again, not a huge effort

AlbertoBrusa (approved):

packages/modular-scripts/src/addPackage.ts
```diff
@@ -157,10 +162,14 @@ async function addPackage({
 
   try {
```
sgb-io, line 163: Are these `console.log` statements intentional?

sgb-io (approved):
LGTM, just 1 comment about some possible debug artefacts

PR #2238, opened 2022-12-19
Sunset modular-site 

AlbertoBrusa:


sgb-io (approved):
On the basis that it's not being used and nobody is being pointed at it (to our knowledge), I think this is safe to remove.

PR #2241, opened 2022-12-21
Remove requirement for Apps to be private in modular check

AlbertoBrusa:
Although we believe that Modular packages of type `App` should not be published to NPM, we don't want to prevent users from changing it should they so choose.
The default will still be `"private": true` on new packages of type `App`

### DIFF
--- a/.changeset/purple-numbers-camp.md
+++ b/.changeset/purple-numbers-camp.md
@@ -1,0 +1,5 @@
+---
+'modular-scripts': minor
+---
+
+App type modular packages are no longer required to be private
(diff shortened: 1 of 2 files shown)


ayseenver (approved):

benpryke (approved):

cristiano-belloni (changes requested):
Is there a defined, justified use case for this? If not, why do we want to merge this?

AlbertoBrusa:
> 

We discussed this and, although we believe the standard use case will be for apps to be set as private, and as such that's what we'll set as a default, the general consensus is that we shouldn't completely restrict users should they want to change this for any reason.

cristiano-belloni (approved):

PR #2255, opened 2023-01-13
Document supported CRA features

cristiano-belloni:
As per title - also renaming "Know limitations" to "Developing with ESM Views"


AlbertoBrusa:
Looks good to me - I wonder if we should title it and word it as differences between Modular and CRA rather than just supported features - for developers looking to understand why to use us over plain CRA?

AlbertoBrusa (approved):

sgb-io (approved):

PR #2259, opened 2023-01-16
Unify modular test command interface

cristiano-belloni:
- `modular test` now accepts package names as arguments, and regexes behind the `--regex` option flag
- `modular test` now accepts any combination of `[packages...]`, `--regex`, `--changed`, `--ancestors`, `--descendants`
- `modular build` builds all packages when invoked without selective arguments / options

- [x] Remove `--package` and accept it as an argument
- [x] Create `--regex` and accept it as a variadic option
- [x] Use `selectWorkspaces` to unify interface
- [x] Merge regexes discovered with packages
- [x] `yarn build` with no args builds all packages
  - [x] Tests for build
  - [x] Docs for build
- [x] `yarn test` verbose log
- [x] Check that `yarn test` without selective args has the same behaviour of Modular 3
- [x] Write docs
- [x] Improve `test.test.ts`

### DIFF
--- a/docs/commands/test.md
+++ b/docs/commands/test.md
@@ -36,9 +36,9 @@ identified test files will be run in non-predictable order. When `packages` is
 empty and no selective options have been specified (for example when running
 `yarn modular test`), all tests in the monorepo will be executed. When
 `packages` contains one or more non-existing package name, the non-existing
-packages will be ignored without an error. If any package or selective options
-has been defined but the final set of regular expressions is empty, Modular will
-write a message to `stdout` and exit with code `0`.
+packages will be ignored without an error. If any package or selective option
+have been defined but the final set of regular expressions is empty, Modular
+will write a message to `stdout` and exit with code `0`.
 
 ### Modular-specific options
 
--- a/docs/commands/test.md
+++ b/docs/commands/test.md
@@ -26,14 +26,14 @@ This contains the setup for tests corresponding to
 ### Arguments
 
 `[packages ...]`: Packages is a list of packages that the user wants to run a
-test on. The specified set of oackages can be agumented by the selective test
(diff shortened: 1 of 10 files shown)


packages/modular-scripts/src/__tests__/test.test.ts
```diff
@@ -195,135 +173,83 @@ describe('Modular test command', () => {
 
     beforeEach(() => {
       // Create random dir
-      randomOutputFolder = tmp.dirSync({ unsafeCleanup: true }).name;
+      randomOutputFolder = createModularTestContext();
```
AlbertoBrusa, line 176: Do these tests need a mock modular repo or do they just need an empty temp directory? 
createModularTestContext writes files, copies files, creates symlinks - not huge performance impact but still potentially more than needed here?

AlbertoBrusa (commented):
Looks good, just a couple of comments

packages/modular-scripts/src/__tests__/test.test.ts
cristiano-belloni: modular repo - with empty directory it complains that modular's tsconfig is not found

docs/commands/test.md
```diff
@@ -21,51 +21,55 @@ This contains the setup for tests corresponding to
 This contains the setup for tests corresponding to
 [`jest.config.js#setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array).
 
-## Command line options
+## Command line options and arguments
+
+### Arguments
+
+`[packages ...]`: Packages is a list of packages that the user wants to run a
```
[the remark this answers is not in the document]
cristiano-belloni: Good simplification, done!

AlbertoBrusa (approved):

PR #2268, opened 2023-01-25
Do not throw with non-modular packages and exclude them from selection

cristiano-belloni:
- [x] Make `check` pass in presence of non-modular packages
- [x] Show non-modular packages as `unknown` type in `workspace`
- [x] Do not try to test non-Modular packages 
- [x] Fix wording in `build`
- [x] ~Show correct data in `analyze`~


AlbertoBrusa (approved):
Looks good to me

PR #2275, opened 2023-02-07
Fix test command not parsing jest options correctly

AlbertoBrusa:


packages/modular-scripts/src/test/index.ts
```diff
@@ -114,8 +114,30 @@ async function test(options: TestOptions, packages?: string[]): Promise<void> {
   const additionalOptions: string[] = [];
   const cleanRegexes: string[] = [];
```
AlbertoBrusa, line 115: We could remove the logic to clean the regex options? not 100% confident on if the problem still exists now that it's not the default option - but right now it works and an argument can be made for leaving it as is

cristiano-belloni: I think we don't have to clean the regexes now, but maybe it's a TODO for when we have some time to think about it. Can we factor out the `clean{Regexes|Packages}` functions though? Both seem to be doing the same thing.

PR #2278, opened 2023-02-09
Modular as a task runner - `modular test`

cristiano-belloni:
Modular can run test on non-Modular packages (i.e. the packages that don't have a `modular` field in their `package.json` but have a `scripts.test` field defined). This solution derives a list of packages from the command selection options, discards non-testable packages, then uses Jest for the remaining Modular packages (as before) and finally runs `yarn workspace test` for each testable non-Modular package.

---

- [x] Fix zero coverage :(
- [x] Remove comments in code
- [x] JSDoc
- [x] Tests


ayseenver (approved):

package.json
```diff
@@ -14,7 +14,7 @@
     "lint:fix": "yarn lint --fix",
     "create-modular-react-app": "ts-node packages/create-modular-react-app/src/cli.ts",
     "modular": "ts-node packages/modular-scripts/src/cli.ts",
-    "test": "yarn modular test --watchAll false --runInBand --env node",
+    "test": "yarn modular test --regex --watchAll false --runInBand --env node",
```
sgb-io, line 17: How come you're adding `--regex` without an argument to this?

packages/modular-scripts/src/__tests__/unobtrusiveModular.test.ts
```diff
@@ -17,6 +17,7 @@ describe('When there is a non-Modular package with a build script', () => {
 
   beforeAll(() => {
     tempModularRepo = createModularTestContext();
+    console.log(tempModularRepo);
```
sgb-io, line 20: Intentional to leave this in?

AlbertoBrusa, line 20: Did you mean to leave this in?

AlbertoBrusa (approved):
Looks good

packages/modular-scripts/src/__tests__/unobtrusiveModular.test.ts
benpryke: Should this still be in here?

package.json
cristiano-belloni: Modular is "special": it needs to be tested with Modular but wants to be built in a custom way. `--regex` will be passed to Jest, which will test everything that matches its patterns in every package (empty string works like `*`), which, in this repo, contains non-modular tests too. This is a necessary trick until we are ready to build Modular with Modular.

packages/modular-scripts/src/__tests__/unobtrusiveModular.test.ts
cristiano-belloni: Removed :)

cristiano-belloni: Removed :)

PR #2309, opened 2023-03-15
modular start: fix and enhance https support in esbuild mode

sgb-io:
Previously, running `start` in esbuild mode with `HTTPS=true` led to an SSL protocol error. The start command for esbuild did not have the full wiring for HTTPS support.

- Fix: Fixed the dev server in esbuild mode to properly enable HTTPS, including a default generated-and-cached self-signed certificate, so that providing a custom certificate is not required by default. This is **consistent with the webpack behaviour**; the self-signed certificate generation code is the same and we store it in the same cache location.
- Enhancement: Enables the [use of custom certificates](https://create-react-app.dev/docs/using-https-in-development/#custom-ssl-certificate) in esbuild mode
- Fix: Made esbuild and webpack mode look for SSL cert files in a consistent way: the individual package directory, then the modular root. Previously, in webpack mode, it would only look for cert files in the individual package directory.
- Internal: Update the esbuild start command to exit in the same way as the webpack one

TODO

- [x] Point at appropriate target branch, add changeset
- [x] Test e2e internally with a custom certificate / ca
- [x] Update the websocket server to accept the appropriate server at the appropriate time
- [x] Debug a reported difference in how the SSL file paths are treated in webpack vs esbuild
- [x] Debug/fix broken esm view tests
- [x] See if a puppeteer test can cover this use-case


packages/modular-scripts/src/esbuild-scripts/start/utils/generateSelfSignedCert.ts
```diff
@@ -0,0 +1,144 @@
...
+// This code is mostly duplicated from webpack-dev-server, with some typings added.
+// To provide consistent https behaviour in esbuild mode, we use the same directories to store the generated cert.
+// See: https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js
+
+function findCacheDir() {
+  const cwd = process.cwd();
+
+  let dir: string | undefined = cwd;
+
+  for (;;) {
```
cristiano-belloni, line 20: This loop finds the first parent package.json - I believe we already use `find-up` for this in other parts of the code -  can we do that here as well for simplicity / readability? That infinite loop looks scary.

sgb-io: As per the comment near the top - this file is >85% duplicated from webpack-dev-server: https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L418

The only changes I'm making are

- Types
- Necessary material changes that allow us to achieve the goal of this PR

The infinite loop here is just duplicated. We can diverge and use our own implementation, but is that we want?

packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -143,13 +146,55 @@ class DevServer {
     await this.startEsbuildServer();
     await this.hostRuntime();
 
+    const { SSL_CRT_FILE, SSL_KEY_FILE, HTTPS } = process.env;
+    const isHttps = HTTPS === 'true';
+
+    let selfSignedCert: Buffer | undefined;
+    if (isHttps) {
+      selfSignedCert = await generateSelfSignedCert();
```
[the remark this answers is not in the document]
sgb-io: I've had to restructure things a bit in 8c8e9c1190730acf8356d7a8a9db47cbd75047b7 after realising that the websocket server also needs to be booted with the correct server instance (whereas before, it would boot the websocket server without https, right away).

I have resolved this particular inefficiency as part of that change.

packages/modular-scripts/src/esbuild-scripts/start/index.ts
```diff
@@ -139,30 +137,64 @@ class DevServer {
...
-        // eslint-disable-next-line @typescript-eslint/no-misused-promises
-        this.server = this.express.listen(this.port, this.host, async () => {
-          await openBrowser(this.urls.localUrlForBrowser);
-          resolve(this);
+    const { SSL_CRT_FILE, SSL_KEY_FILE, HTTPS } = process.env;
+    const isHttps = HTTPS === 'true';
+
+    if (isHttps) {
+      // By default, use a self-signed, generated cert
+      const selfSignedCert = await generateSelfSignedCert();
```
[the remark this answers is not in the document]
sgb-io: This is resolved as of 69e8248f9813b4425bac231544fc5d778c220377

packages/modular-scripts/src/esbuild-scripts/start/utils/generateSelfSignedCert.ts
AlbertoBrusa: Sounds like we can de-duplicate some of this since my PR Types webpack-dev-server - and maybe improve this logic too

sgb-io:
Directly replaced by https://github.com/jpmorganchase/modular/pull/2310 which is a rebased version of this work

PR #2283, opened 2023-02-16
Webpack TS Configuration Initial Implementation

AlbertoBrusa:
**Background**
Webpack now supports configuration written in TypeScript.
Currently, modular-scripts has two folders of JS code copied over from CRA, adapted as needed, to create the Webpack configuration for Modular. Modular's build and start script run JS scripts with execa() to start and build with Webpack.
Problems: 
- JavaScript Code outside the /src/ directory - hard to maintain and work on 
- Duplicated code - some of these scripts have been translated to TS and placed in /src/ as they are required by other parts of Modular 
- Running the scripts with execa() means having to pass all variables as Environment variables

**This PR**
We now create the Webpack configuration in TypeScript, and run the build and start scripts as functions, passing variables directly.
- No more internal Env variables 
- TypeChecking in Webpack configuration 
- Moved all scripts into /src/build-scripts where we can manage them better
- react-scripts and react-dev-utils are now integrated in the modular build logic under build-scripts/webpack-scripts/ and match what esbuild-scripts do/how they are called
- Duplicated logic from esbuild and webpack now deduped and placed in common-scripts

To Do: 

- [x] Fix CDN rewriting cause it seems very much broken right now (esmView.test.ts)
- [x] Fix conflicts with https://github.com/jpmorganchase/modular/pull/2282
- [x] Improve folder structure 
- [x] Refactor/rename/cleanup 

There's still a lot of work to be done in future PRs:
- Still some duplicated pieces of code
- Many scripts still in JS
    - The script that provides the baseConfig for Webpack has a non-typed dependency, and in turn it uses a lot of scripts that have to stay written in JS 

**Few notes for those reviewing this PR**:
Things have been moved around so much that a lot of code appears as 'new', while in fact it's just moved/typed but using existing logic. I've held back from making actual functional code changes to existing scripts unless absolutely necessary.

PR #2310, opened 2023-03-17
modular start: fix and enhance https support in esbuild mode (rebased)

sgb-io:
__Direct replacement for https://github.com/jpmorganchase/modular/pull/2309, but rebased/refactored to work in @AlbertoBrusa's branch__

Previously, running `start` in esbuild mode with `HTTPS=true` led to an SSL protocol error. The start command for esbuild did not have the full wiring for HTTPS support.

- Fix: Fixed the dev server in esbuild mode to properly enable HTTPS, including a default generated-and-cached self-signed certificate, so that providing a custom certificate is not required by default. This is **consistent with the webpack behaviour**; the self-signed certificate generation code is the same and we store it in the same cache location.
- Enhancement: Enables the [use of custom certificates](https://create-react-app.dev/docs/using-https-in-development/#custom-ssl-certificate) in esbuild mode
- Fix: Made esbuild and webpack mode look for SSL cert files in a consistent way: the individual package directory, then the modular root. Previously, in webpack mode, it would only look for cert files in the individual package directory.
- Internal: Update the esbuild start command to exit in the same way as the webpack one


sgb-io:
Directly replaced by https://github.com/AlbertoBrusa/modular/pull/2, which will merge the work into that fork, and be merged eventually via https://github.com/jpmorganchase/modular/pull/2283

PR #2315, opened 2023-03-20
Selective `modular lint`

cristiano-belloni:
Modular lint accepts selective options and now doesn't die (but defaults to `--all`) if called on a directory that's not a Git repository. This closes https://github.com/jpmorganchase/modular/pull/2208

- [x] Selective lint
- [x] Lint tests
- [x] Fork eslint config
- [x] Changeset
- [x] Docs

### DIFF
--- a/docs/commands/lint.md
+++ b/docs/commands/lint.md
@@ -5,24 +5,55 @@ title: modular lint
 
 # `modular lint [options] [regexes...]`
 
-`modular lint` will check the diff between the current branch and your remote
-origin default branch (i.e. `master` or `main`) and only lint the `.ts`, `.tsx`,
-`.js`, `.jsx` files that have changes.
+`modular lint` without arguments or options will check the diff between the
+current branch and your remote origin default branch (i.e. `master` or `main`)
+and only lint the `.ts`, `.tsx`, `.js`, `.jsx` files that have changes. When in
+CI or when `--all` is provided, it will default to lint the entire codebase.
+Wneh `--staged` is provided, it will only lint the files staged on git.
+
+Alternatively, when it is invoked with one or more selective options
+(`--packages`, `--ancestors`, `--descendants` or `--changed`), it will lint all
+the source code files in the provided `--packages`, optionally agumenting the
+set of packages by adding their ancestors (`--ancestors`), their descendants
+(`--descendants`) or the files that have changed (`--changed`), calculated
+comparing the current state of the git repository with the branch specified by
+`compareBranch` or, if `compareBranch` is not set, with the default branch.
 
 It uses your project eslint config to lint and jest runner to report on the lint
 status of each file (`PASS` or `FAIL`). If a file has a lint warning, the
 command will be marked as failed, with printed eslint errors for each
 warning/error.
 
-When in CI, it will default to lint the entire codebase. When not in CI, it will
-default to only lint the diffed files.
-
 ## Options:
 
-`--all`: Lints the entire codebase
+`--all`: Lints the entire codebase (not compatible with `--staged` and with the
+Selective Options)
 
-`--staged`: Lints only files staged on git (not compatible with `--all`)
+`--staged`: Lints only files staged on git (not compatible with `--all` and with
+the Selective Options)
 
 `--fix`: Allows eslint to fix the errors and warnings that do not require manual
 intervention wherever possible. Restages any fixed files that were previously
 staged when used in combination with `--staged`.
+
+### Selective Options
+
+`--packages [packages...]`: A list of one or more packages to lint. Can be
+combined with all the other selective options.
+
+`--ancestors`: Take the packages specified via `--packages` or `--changed` and
+add their ancestors (i.e. the packages that have a direct or indirect dependency
+on them) to the lint list.
+
+`--descendants`: Take the packages specified via `--packages` or `--changed` and
+add their descendants (i.e. the packages they directly or indirectly depend on)
+to the lint list.
+
+`--changed`: Add all the packages whose workspaces contain files that have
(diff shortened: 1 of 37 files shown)


docs/commands/lint.md
```diff
@@ -5,24 +5,55 @@ title: modular lint
...
 # `modular lint [options] [regexes...]`
 
-`modular lint` will check the diff between the current branch and your remote
-origin default branch (i.e. `master` or `main`) and only lint the `.ts`, `.tsx`,
-`.js`, `.jsx` files that have changes.
+`modular lint` without arguments or options will check the diff between the
+current branch and your remote origin default branch (i.e. `master` or `main`)
+and only lint the `.ts`, `.tsx`, `.js`, `.jsx` files that have changes. When in
+CI or when `--all` is provided, it will default to lint the entire codebase.
+Wneh `--staged` is provided, it will only lint the files staged on git.
```
sgb-io, line 12:
```suggestion
When `--staged` is provided, it will only lint the files staged on git.
```

docs/commands/lint.md
```diff
@@ -5,24 +5,55 @@ title: modular lint
...
 command will be marked as failed, with printed eslint errors for each
 warning/error.
 
-When in CI, it will default to lint the entire codebase. When not in CI, it will
-default to only lint the diffed files.
-
 ## Options:
 
-`--all`: Lints the entire codebase
+`--all`: Lints the entire codebase (not compatible with `--staged` and with the
```
sgb-io, line 29: Related, do you think typecheck should have the `--all` option, too?

docs/commands/lint.md
```diff
@@ -5,24 +5,55 @@ title: modular lint
...
+(`--descendants`) or the files that have changed (`--changed`), calculated
+comparing the current state of the git repository with the branch specified by
+`compareBranch` or, if `compareBranch` is not set, with the default branch.
 
 It uses your project eslint config to lint and jest runner to report on the lint
 status of each file (`PASS` or `FAIL`). If a file has a lint warning, the
 command will be marked as failed, with printed eslint errors for each
 warning/error.
 
-When in CI, it will default to lint the entire codebase. When not in CI, it will
```
sgb-io, line 17: I guess you removed this because it depends on what selective options they use (or none at all). However, without looking at the code, I'm guessing that `!!CI` mattered: this is also the case in typecheck. If I'm right, that behaviour needs to be preserved, right?

packages/modular-scripts/src/lint.ts
```diff
@@ -60,6 +126,8 @@ async function lint(
     generateJestConfig(jestEslintConfig),
   ];
 
+  // console.log({ testArgs });
```
sgb-io, line 129: probable unintended `console.log`

docs/commands/lint.md
sgb-io: (Assuming that behaviour *is* preserved, I guess it should be described here)

packages/modular-scripts/src/__tests__/lint.test.ts
```diff
@@ -97,3 +101,81 @@ describe('Modular lint', () => {
...
+      (acc) => `${acc}..${path.sep}`,
+      '',
+    ),
+    '__fixtures__',
+    'selective-lint',
+  );
+
+  let randomOutputFolder: string;
+
+  beforeEach(() => {
```
sgb-io, line 118: Consider using non-execa (i.e. not relying on `runModularPipeLogs` and instead calling `lint()` directly) techniques like in the selective typecheck PR? Should be a bit faster, and test coverage should get tracked properly.

I think it's viable for `lint`, just a case of mocking the modular root after creating the temp modular repo. The typecheck example of this approach [is here](https://github.com/jpmorganchase/modular/blob/18bb7a9c31854e47910fe2c1d98ab39d28dda6ad/packages/modular-scripts/src/__tests__/typecheck.test.ts)

docs/commands/lint.md
cristiano-belloni: No. `--all` is here only because removing it would be breaking. Typecheck does the right thing already (typechecks the whole project when no options are specified)

docs/commands/lint.md
cristiano-belloni: I removed it because it was misleading and incomplete (it ignored `--regex` and `--staged`). I'll rewrite it.

packages/modular-scripts/src/__tests__/lint.test.ts
cristiano-belloni: Created a separate spike ticket for this.

docs/commands/lint.md
cristiano-belloni: Done.

packages/modular-scripts/src/__tests__/lint.test.ts
```diff
@@ -97,3 +101,81 @@ describe('Modular lint', () => {
...
   });
 });
+
+describe('lint command can successfully do selective tests based on selected packages', () => {
+  const fixturesFolder = path.join(
+    __dirname,
+    Array.from({ length: 4 }).reduce<string>(
+      (acc) => `${acc}..${path.sep}`,
+      '',
+    ),
```
benpryke, line 111: This will always resolve to `'../../../../'`. Why not simply include the raw string?

sgb-io (commented):
Rest LGTM

sgb-io (approved):

packages/modular-scripts/src/__tests__/lint.test.ts
cristiano-belloni: Because `path.sep` is different in Linux | Windows | MacOS

PR #2318, opened 2023-03-21
Yarn > 1 Support

AlbertoBrusa:
CMRA Changes: 
- Add support for Yarn 3 
- Error on Yarn 2

Added documentation specifying how we support various versions of Yarn 

Modify GitHub Test matrix to now test with Yarn 3


docs/compatibility.md
```diff
@@ -10,16 +10,36 @@ Modular is based on
...
+Yarn v2 is not supported by Modular, due to
+[PnP](https://yarnpkg.com/features/pnp) and other issues.
+
+### Yarn v3
+
+Yarn v3 is partially supported; we are actively trying to support more features.
+
+All Modular commands and functionality should work as expected.
+
+The project's `.yarnrc.yml` must include the `nodeLinkler` property set to
```
cristiano-belloni, line 26: Is this `nodeLinker` ?

AlbertoBrusa: thanks good catch

AlbertoBrusa:
> This PR has 89 commits - I guess you didn't checkout this from `main` and it's trying to replay all the commits from your ts PR? Just for clarity's sake, since it's mergeable, can you squash it or rebase it? It won't make any difference in the end since we squash it anyway, so if you prefer to just edit the message when we squash merge it's ok for me.

Yep my bad - I squash commit and edit into a clear set of changes when there are too many commits anyways, so will do that

PR #2320, opened 2023-03-21
Implement cascade configuration

AlbertoBrusa:
Allows users to specify a root config that's overriden by package-specific configurations 
Added debug logging to allow users to see what configured values are used and from where when verbose/Debug logging is on


cristiano-belloni (approved):

packages/modular-scripts/src/__tests__/utils/config.test.ts
```diff
@@ -13,28 +13,45 @@ const configFixtures = path.join(modularRoot, '__fixtures__', 'test-config');
...
     const result = runModularPipeLogs(
       tempModularRepo,
-      `build test-app --verbose`,
+      `build ${app} --verbose`,
       'true',
     );
     expect(result.stdout).toContain('Building with esbuild');
     expect(result.exitCode).toBe(0);
   });
+  it('builds using webpack if a package specific config gile overrides the root config', () => {
```
ayseenver, line 38: I'm not sure if this is really testing the case where a package is overwriting the global config, since you're only importing the webpack config. Maybe it would be better to rename the config files to "root.config.js" and "package.config.js", import both and check that the package one is being picked up only?

AlbertoBrusa: So in this specific test you have:
-  .modular.js in the root specifying to use esbuild (added in the beforeEach code)
- .modular.js in the package specifying use webpack 

The test is, if it builds with webpack, the package config has overridden the root config (in the previous test case, there is only a root config, and we check that it builds with esbuild)

I feel like this covers it? unless I'm misunderstanding what you're saying

ayseenver: oh I think you're right! The base config still exists from the previous test and this is overwriting it

AlbertoBrusa: Yep the root config is added in all 3 tests thought the beforeEach

PR #2321, opened 2023-03-22
Implement selective typecheck

sgb-io:
## Selective TypeCheck

- Implements selective typecheck, i.e adds the `--ancestors`, `--descendants`, `--changed` and `--compareBranch` options to `modular typecheck`
- Maintains backwards compatibility with the existing `modular typecheck`
- Adds new docs that describe how Modular currently treats `tsconfig.json` files, plus docs for the new options
- Updates the tests for `typecheck` to use a new approach, and cover new cases

## New unit testing approach: calling commands directly

A side mission during this PR was to call `typecheck()` directly from test code, instead of rely on `execa` as is very common in the project. The `execa` approach is (a) slow and (b) breaks the tracking of test coverage. `typecheck.test.ts` demonstrates a new approach:

- Call the command directly, avoiding the need for a child process, and fixing jest's ability to track test coverage
- Mock `getModularRoot` in tests, which is used extensively throughout the project
- Create a secondary `getModularRoot` clone called `getRealModularRootInTest` that is designed to be used in test code, i.e. avoid any mocks that might be applicable
- Avoid the `actionPreflightCheck`, which complicates the mocking (it contains a call to `getModularRoot`) and slows things down. Plus, this check has no value in test (it can be tested in isolation).

It should be noted that I've used `jest.doMock` which avoids jest's [default hoisting behaviour](https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options). In other words, it lets us mock the modular root just before importing and calling a modular command. To work with this approach, I dynamically import the `typecheck` command. It may also be possible to use a more static setup.

Not only is this approach faster and simpler, the coverage for the test command has gone from `0` to `>90%`.

TODO

- [x] New docs
- [x] Repoint PR to v4.3 or main when v4.2 is released
- [x] Tests
- [x] Changeset

### DIFF
--- a/packages/modular-scripts/src/typecheck.ts
+++ b/packages/modular-scripts/src/typecheck.ts
@@ -5,34 +5,106 @@ import getPackageMetadata from './utils/getPackageMetadata';
 import * as logger from './utils/logger';
 import getModularRoot from './utils/getModularRoot';
 import actionPreflightCheck from './utils/actionPreflightCheck';
+import { getAllWorkspaces } from './utils/getAllWorkspaces';
+import { selectWorkspaces } from './utils/selectWorkspaces';
 
-async function typecheck(): Promise<void> {
-  const { typescriptConfig } = await getPackageMetadata();
+import type { JSONSchemaForTheTypeScriptCompilerSConfigurationFile as TSConfig } from '@schemastore/tsconfig';
 
-  const { _compilerOptions, ...rest } = typescriptConfig;
+type CompilerOptions = TSConfig['compilerOptions'];
 
-  const tsConfig: typeof typescriptConfig = {
+export interface TypecheckOptions {
+  ancestors: boolean;
+  descendants: boolean;
+  changed: boolean;
+  compareBranch: string;
+}
+
+const COMPILER_OPTIONS_ALLOW_LIST: string[] = ['jsx'];
+const DEFAULT_COMPILER_OPTIONS: CompilerOptions = {
+  noEmit: true,
+};
+const DEFAULT_EXCLUDES: string[] = [
+  'node_modules',
+  'bower_components',
+  'jspm_packages',
+  'tmp',
+  '**/dist-types',
+  '**/dist-cjs',
+  '**/dist-es',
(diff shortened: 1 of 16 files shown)


packages/modular-scripts/src/typecheck.ts
```diff
@@ -5,35 +5,107 @@ import getPackageMetadata from './utils/getPackageMetadata';
...
+
+  const combinedCompilerOptions: CompilerOptions = {
+    ...DEFAULT_COMPILER_OPTIONS,
   };
 
-  // If compilerOptions.jsx is defined use it otherwise, depend on the extended config by not setting it to undefined.
-  if (tsConfig.compilerOptions && rest.compilerOptions?.jsx) {
-    tsConfig.compilerOptions.jsx = rest.compilerOptions.jsx;
+  // Where the user has defined allowlist-supported compiler options, apply them
+  for (const item of COMPILER_OPTIONS_ALLOW_LIST) {
```
sgb-io, line 57: This block is currently uncovered, worth adding a test case for it (essentially, using the jsx option)

packages/modular-scripts/src/utils/getPackageMetadata.ts
```diff
@@ -56,9 +56,10 @@ async function getPackageMetadata() {
   // validate tsconfig
   // Extract configuration from config file and parse JSON,
   // after removing comments. Just a fancier JSON.parse
+  const userTsConfigPath = path.join(modularRoot, typescriptConfigFilename);
```
sgb-io, line 59: This is a bugfix. The second argument was previously reading `tsconfig.json` from the `cwd()`, since no modular root was being used. This is not consistent with the first argument, where the modular root is used.

That said, in practice, there was no actual way for users to be affected by this bug - since commands are supposed to be run in the root already (so not using the modular root resulted in the correct file being used), and preflight errors actually get in the way first if you try to run something like `typecheck` from the non-root dir.

The bugfix here is needed so that we can effectively test the code in test, i.e. use a faked modular root and have it respected.

PR #2333, opened 2023-03-30
Implement integration tests

AlbertoBrusa:
Github Action that runs integration test bash scripts that set up a verdaccio instance, build and publish all modular packages and templates, and runs through setting up a new repo with CMRA, and runs all modular commands on it for each template we provide.


integration-test-scripts/verdaccio-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+plugins: ./plugins
+
+web:
+  title: Verdaccio
+
+auth:
+  htpasswd:
+    file: ./htpasswd
+
+# Back Verdaccio onto Artifactory instead of NPM
```
sgb-io, line 13: This comment should be removed (no usage of artifactory here)

AlbertoBrusa: Good spot, thanks

PR #2337, opened 2023-04-03
`select` command

cristiano-belloni:
- [x] Write command
- [x] Refactor `selectWorkspaces` to handle parallel building
- [x] Check why root package is included in unordered response
- [x] Tests
- [x] Docs


packages/modular-scripts/src/__tests__/select.test.ts
```diff
@@ -0,0 +1,398 @@
...
+      cwd: tempModularRepo,
+    });
+
+    execa.sync('git', ['commit', '-am', '"First commit"'], {
+      cwd: tempModularRepo,
+    });
+  });
+
+  it('selects nothing when everything committed', () => {
+    const result = runModularPipeLogs(tempModularRepo, 'select --changed');
```
AlbertoBrusa, line 55: Is it possible to run the select command and mock getModularRepo as with the esmView tests instead of using runModularPipeLogs? I feel like we should apply the new approach to new tests instead of expanding the use of execa 
We haven't replaced tests that pipe logs to check stdout/err yet so it will require some thinking about

AlbertoBrusa (commented):
Looks good - the tests could possibly be more efficient using the new approach but its debatable whether we start doing it as we go along  or just leave it all as part of the separate bulk effort

packages/modular-scripts/src/__tests__/select.test.ts
cristiano-belloni: I'll capture this in a separate ticket - trying to keep the PRs single-themed.

cristiano-belloni: Done.

benpryke (approved):

PR #2341, opened 2023-04-13
`modular build` builds in concurrent batches

cristiano-belloni:
Enables batched algorithms for `modular build` with a custom max concurrency level. Build jobs that can be parallelised are executed in consecutive `--concurrencyLevel`-long job parallel batches, that are run one after the other until exhaustion. `--concurrencyLevel` level is a new user option that defaults to the number of logical CPUS in the machine.

This also fixes a bug in `modular select` where the parallel dependency array can sometimes have holes.

---

- [x] Implement batched algo
- [x] Tests
- [x] Docs

### DIFF
--- a/packages/modular-scripts/src/utils/options.ts
+++ b/packages/modular-scripts/src/utils/options.ts
@@ -18,7 +18,7 @@ export function validateCompareOptions(
 }
 
 /**
- * Validates and conver concurrency option to number.
+ * Validates and converts concurrency option to number.
  *
  * Throws and exits when invalid value is detected.
  */
--- a/packages/modular-scripts/src/utils/options.ts
+++ b/packages/modular-scripts/src/utils/options.ts
@@ -27,6 +27,7 @@ export function computeConcurrencyOption(
 ): number {
   if (!concurrencyLevel) {
     // If concurrencyLevel is not supplied, default to the number of CPUs. If the number of CPUs is 0 (/proc or equivalent is unavailable), default to no concurrency.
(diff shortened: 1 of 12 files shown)


AlbertoBrusa (approved):

packages/modular-scripts/src/utils/options.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (!concurrencyLevel) {
+    // If concurrencyLevel is not supplied, default to the number of CPUs. If the number of CPUs is 0 (/proc or equivalent is unavailable), default to no concurrency.
+    return os.cpus().length || 1;
+  }
+
+  // Otherwise, try to parse it
+  const concurrency = parseInt(concurrencyLevel, 10);
+
+  // If not a number or a negative number, bail out with an error.
+  if (isNaN(concurrency) || concurrency < 0) {
```
sgb-io, line 37: What happens if someone provides a big number? Let's say they typo a couple zeros on the end and put `600`. Is this fine or do we need an upper boundary?

cristiano-belloni: > What happens if someone provides a big number? Let's say they typo a couple zeros on the end and put `600`. Is this fine or do we need an upper boundary?

It's fine, because:
1) in practice, the number of things a project can build concurrently at a given stage will be much less than 600 (even if the project has 600 packages, they will be probably interdependent)
2) the worse that can happen is that Modular goes out of memory, in which case the user will just correct their command

PR #2343, opened 2023-04-13
Enhance tests Pt. 1 

AlbertoBrusa:


cristiano-belloni (approved):

sgb-io:
> Updated addPackage.test.ts to not use execa where possible - couldn't do that where 'path' was passed to addPackage as some path.relative() logic breaks due to it using CWD

As in, contains code referencing `process.cwd()`? That can also be manually mocked, if it helps.

This change looks good though, decent improvement 👍

AlbertoBrusa:
> > Updated addPackage.test.ts to not use execa where possible - couldn't do that where 'path' was passed to addPackage as some path.relative() logic breaks due to it using CWD
> 
> As in, contains code referencing `process.cwd()`? That can also be manually mocked, if it helps.
> 
> This change looks good though, decent improvement 👍

So the code uses path.relative() which is relative to the CWD - I don't know if mocking process.cwd would affect that - And anyways we can't mock it for the same reason we can't mock path.relative - addPackage needs to know the path of the temp modular repo relative to the *actual* cwd for when it tries to run yarn commands on it  (not sure why it doesn't just always use absolute paths - I could go down that road changing things if we think its worth it

PR #2352, opened 2023-04-18
Enable "strict" compiler option

cristiano-belloni:
- Enable `strict`
- Support falsy values in configuration
- Slightly better typing

### DIFF
--- a/packages/modular-scripts/src/typecheck.ts
+++ b/packages/modular-scripts/src/typecheck.ts
@@ -19,7 +19,7 @@ export interface TypecheckOptions {
   compareBranch: string;
 }
 
-const COMPILER_OPTIONS_ALLOW_LIST: string[] = ['jsx'];
+const COMPILER_OPTIONS_ALLOW_LIST: string[] = ['jsx', 'strict'];
 const DEFAULT_COMPILER_OPTIONS: CompilerOptions = {
   noEmit: true,
 };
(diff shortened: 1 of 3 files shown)


packages/modular-scripts/src/typecheck.ts
```diff
@@ -55,7 +61,7 @@ function restrictUserTsconfig(
 
   // Where the user has defined allowlist-supported compiler options, apply them
   for (const item of COMPILER_OPTIONS_ALLOW_LIST) {
-    if (compilerOptions && compilerOptions[item]) {
+    if (compilerOptions?.[item] !== undefined) {
```
sgb-io, line 64: nit: I get that this is needed to allow `false`, but by explicitly checking for `!== undefined` I think you might be technically allowing something like `null`. Saying that, it's probably OK to consider that something end users should fix though, rather than validating values here or anything.

sgb-io (approved):

AlbertoBrusa (approved):

packages/modular-scripts/src/typecheck.ts
cristiano-belloni: @sgb-io I guess if we end up with `null` there, it means that the user explicitly (and probably erroneously) ended up with a `null` option in their configuration. I think we should let tsconfig validate and eventually throw in that case, instead of ignoring the option, because it could be a symptom of something wrong (i.e. buggy conf script / json from the user). `undefined` is a special case, because unfortunately it means both "absence of key" and "key was set to undefined" in Javascript, and there's no way of distinguishing that if the conf comes from outside.

benpryke (approved):

PR #2362, opened 2023-04-25
Expand CDN template

cristiano-belloni:
> - `[path]` is replaced with the version of the submodule path in the `import`
  statement, if present. For example, in the statement
  `import @scope/name/path/to/module`, `[path]` in the CDN template will be
  replaced with `/path/to/module` (initial separator included), while in the
  statement `import @scope/name`, `[path]` will be replaced by the empty string.
  If the `[path]` substring is not specified in the CDN template but the subpath
  module is present in the `import` statement, the submodule path will be added
  ad the end of the template by default.
  
- [x] Rewrite ESM Views to generate the dependency name in the bundler plugin
- [x] Expand template to use other subsets
- [x] Tests
- [x] Test manually with CSS import
  - [x] esbuild
  - [x] Webpack
- [x] Docs


cristiano-belloni:
If we add `[path]`, this changes becomes breaking, so this was moved to 5.0

docs/esm-views/esm-cdn.md
```diff
@@ -47,6 +47,14 @@ These are the substrings that are replaced in the template:
...
 - `[resolution]` is replaced with the version of the imported dependency as
   extracted from the yarn lockfile (`yarn.lock`).
+- `[path]` is replaced with the version of the submodule path in the `import`
+  statement, if present. For example, in the import statement
+  `import @scope/name/path/to/module`, the substring `[path]` in the CDN
+  template will be replaced with `/path/to/module` (initial separator included),
+  while in the statement `import @scope/name`, `[path]` will be replaced by the
+  empty string. If the `[path]` substring is not specified in the CDN template
+  but the subpath module is present in the `import` statement, the submodule
+  path will be added ad the end of the template by default.
```
AlbertoBrusa, line 57: typo

AlbertoBrusa (approved):
Looks good to me, just one small typo in the docs

PR #2384, opened 2023-05-10
Polish unobtrusive modular

AlbertoBrusa:
Unobtrusive lint & typecheck by default 
Standardised modular commands to all use the same interface


packages/modular-scripts/src/lint.ts
```diff
@@ -15,121 +14,128 @@ import {
...
-  );
-
-  let lintRegexes = isSelective || all ? processedPackageRegexes : [];
-
-  if (all && isSelective) {
-    logger.warn(
-      'You specified --all already; selective options are redundant.',
+  if ((staged || diff || userRegexes.length) && isSelective) {
+    logger.error(
+      "Can't specify --staged --diff --regex along with selective options",
```
sgb-io, line 50: Slightly more explicit:

```suggestion
      "Can't specify --staged, --diff or --regex along with selective options",
```

docs/commands/lint.md
```diff
@@ -27,38 +31,31 @@ warning/error.
...
 ## Default behaviour
 
-`modular lint` without arguments or options will:
-
-- default to lint the `git diff` between the current branch and the remote
-  origin's default branch when not in CI
-- default to lint everything (essentially like `--all`) when in CI
+`modular lint` without arguments or options will lint all packages in the
+repository, and run any `lint` scripts specified in the `package.json` of any
+non-modular packages.
```
sgb-io, line 36: I think we need some explicit content somewhere in the docs that talks about the selective options, what they are, why they are useful, but also why things can be slightly different between commands.

It might make sense to add a new page or section somewhere else and have both `test` and `lint` point to it.

I think the content would approximately consist of: 'these are the selective options which allow you to narrow down which packages or files are applicable when running X' and 'this is why lint is different to test, this is why start doesn't have them'.

Let me know your thoughts - happy to help write the content

AlbertoBrusa: I'd rather not add yet another file to the documentation - I want documentation to be informational but also practical and concise so that important info doesn't get 'lost'. 

I've re-written the Lint and Test configuration to better bring attention to what options are part of the selective options, that behave the same across all commands, and which are unique to a specific command

PR #2385, opened 2023-05-12
Fix esm-view not starting by downgrading esbuild-loader to 2.x.x

AlbertoBrusa:
- Downgrade esbuild-loader, which fixes starting ESM views in webpack mode. This issue was introduced in https://github.com/jpmorganchase/modular/pull/2332.
- Adds a new test to verify starting ESM views in webpack mode (previously, only the esbuild variant was covered).

Since upgrading esbuild-loader did not actually fix the issue in https://github.com/jpmorganchase/modular/pull/2332, it is a patch change to roll this back. We will separately address the same issue.


sgb-io:
> _ No description provided. _

Suggested description:

> - Downgrade `esbuild-loader`, which fixes `start`ing ESM views in webpack mode. This issue was introduced in #2332.
> - Adds a new test to verify starting ESM views in webpack mode (previously, only the esbuild variant was covered).
> 
> Since upgrading `esbuild-loader` did not actually fix the issue in #2332, it is a patch change to roll this back. We will separately address the same issue.

sgb-io (approved):

Issue #2358, opened 2023-04-20
Docs site is missing a favicon


sgb-io:
It would be nice to have one 😄

Nerajno:
Any specific requirements?

nolanschroeder:
I don't seem to have access to assign this issue to myself. I just forked this project and I will work on it.

PR #2386, opened 2023-05-13
update favicon for docs

ovpv:
#2358


docs/_config.yml
```diff
@@ -36,6 +36,9 @@ search:
 # Supports true (default) or false
 heading_anchors: true
 
+# Set a path/url to a favicon that will be displayed by the browser
+favicon_ico: 'docs/img/modular-hero.ico'
```
ovpv, line 40: @sgb-io  there is a confusion on the relative path for fav icon. should i set it in context of project root or relative root ie. "/img/modular-hero.ico"?

ovpv:
@sgb-io can you please review and approve this?

PR #2391, opened 2023-05-17
Modular respects .gitignore & .modularignore

AlbertoBrusa:
Workspace resolver will respect `.gitignore`, overridden by `.modularignore` if provided

### DIFF
--- a/.changeset/eight-bats-double.md
+++ b/.changeset/eight-bats-double.md
@@ -1,5 +1,6 @@
 ---
 'modular-scripts': major
+'@modular-scripts/workspace-resolver': major
 ---
 
 Modular's workspace resolver now ignores a workspace if covered by a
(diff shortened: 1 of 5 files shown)


.changeset/eight-bats-double.md
```diff
@@ -0,0 +1,6 @@
+---
```
sgb-io, line 1: The workspace resolver package likely needs a changeset for this

AlbertoBrusa: good catch

packages/workspace-resolver/package.json
```diff
@@ -4,8 +4,8 @@
   "license": "Apache-2.0",
   "main": "src/index.ts",
   "dependencies": {
+    "@esm2cjs/globby": "13.1.4",
```
sgb-io, line 7: What was the need to switch to the forked version of `globby`?

AlbertoBrusa: OG globby is now a pure ESM package that we can't import

sgb-io (approved):

PR #2470, opened 2023-10-12
Add modular directory to targetLocations checked by typecheck

ImmanuelBaskaran:
This fixes an issue where the modular folder is ignored when running the typecheck command.

The modular folder in project contains the setupTests.ts file which is used to import packages such as 'testing-library/jest-dom' which extend the types of jest's matchers.

Currently this directory is ignored which causes typecheck to fail if a global import is used in the setupTests.ts


yarn.lock
```diff
@@ -7690,6 +7828,11 @@ jest-docblock@^29.4.3:
   dependencies:
     detect-newline "^3.0.0"
 
+jest-dom@^4.0.0:
```
AlbertoBrusa, line 7831: Did you run yarn again after removing the dependency? seems to have persisted in the lock - i'd revert the yarn lock to its original state

ImmanuelBaskaran: Good catch.

AlbertoBrusa (approved):

reecuepatt-dot (approved):
Os thx

